PR #38, opened 2025-01-09
fix missing ) in example

maxandersen:


mkouba (approved):
Thanks for catching that!

PR #55, opened 2025-01-14
Add a way to select which transport is used when multiple exists

geoand:


core/deployment/src/main/java/io/quarkiverse/mcp/server/deployment/McpBuildTimeConfig.java
```diff
@@ -0,0 +1,17 @@
...
+import io.smallrye.config.ConfigMapping;
+
+@ConfigMapping(prefix = "quarkus.mcp.server")
+@ConfigRoot(phase = ConfigPhase.BUILD_TIME)
+public interface McpBuildTimeConfig {
+
+    /**
+     * If multiple transports extensions are on the classpath, select which one is used
+     */
+    Optional<String> transport();
```
mkouba, line 16: I wonder if this config option still makes sense... given the fact that both transports can coexist side by side?

WDYT @geoand @maxandersen

jmartisk: In the client-server integration tests, I'm currently depending on exposing both at the same time. I don't see why we would restrict to one, maybe it would make more sense to be able to selectively disable each separately if desired.

geoand: If it works, all the better :)

mkouba: Ok, I'm going to close this one for now and we can revisit later if needed.

PR #79, opened 2025-01-21
stdio: redirect console logging to stderr

### LINKED ISSUE #77: enable quarkus.log.console.stderr=true instead of completely remove it
the mcp log protocol does not actually (yet) show up in typical clients. 

i can confirm that other mcp servers send output to stderr and that does not break things.

I can even do it manually in an mcp server in quarkus and things just work.

Thus I would suggest instead of having mcp stdio default disable console.log either send stderr=true or simply require users to set it.

Or at least be able to let users set that property and honor it.

mkouba:
- resolves #77


maxandersen:
i like it !

PR #84, opened 2025-01-23
feat: honor EOF as graceful exit

maxandersen:
this do graceful exit when EOF because when you quit Claude Desktop java processes keeps hanging around as no signal is otherwise sent.

Also stdio spec at https://spec.modelcontextprotocol.io/specification/2024-11-05/basic/transports/ states "Close stdin, terminate subprocess"

Thus I think this is actual a bug/missing feature.


mkouba:
> Also stdio spec at https://spec.modelcontextprotocol.io/specification/2024-11-05/basic/transports/ states "Close stdin, terminate subprocess"

Well, the spec shows in a diagram that it's the client that closes the stdin and terminates subprocesses but I agree that it's reasonable to exit the app when EOF is reached.

PR #88, opened 2025-01-27
core: introduce API to encode return values as response objects

### LINKED ISSUE #85: Add API to encode aritrary objects as Content and ResourceContents
It should be possible to return any Java type from a feature method and this API will make sure the returned object is encoded as the required content type.

By default, returned objects will be converted to JSON.

For example, if there is a Tool method like:

```java
public record MyObject(String name, int score) {}

public static class MyTools {

        @Tool
        MyObject bravo() {
            return new MyObject("foo", 10);
        }
    }
```

Then the returned object is sucessfull `ToolResponse` with `TextContent` and text value `{\"name\":\"foo\",\"score\":10}`.

However, a user can provide a custom encoder like:

```java
@Priority(1) // the default encoder has priority 0
@Singleton
public class MyObjectEncoder<MyObject> {

    @Override
    public boolean supports(Type type) {
        return MyObject.class.equals(type);
    }

    @Override
    public Content encode(MyObject value) {
         return new TextContent(value.toString());
    }
}
```

So that the returned object is sucessfull `ToolResponse` with `TextContent` and text value something like `MyObject[name="foo", score=10]`.

mkouba:
- resolves #85

We introduce several encoders that can be used to encode any return value to a relevant response type. 

Specifically, there is `ToolResponseEncoder` that could be used to encode any return value to a `ToolResponse`; `ContentEncoder` to encode any return value to `Content`; `ResourceContentsEncoder` to `ResourceContents` and `PromptResponseEncoder` to `PromptResponse`. 

We also register `JsonTextContentEncoder` and `JsonTextResourceContentsEncoder` by default. As a result, any "custom" return value returned from a `@Tool` method is encoded as `TextContent` containing the value encoded as JSON (using Jackson), and any "custom" return value returned from a `@Resource` method is encoded as `TextResourceContents` containing the value encoded as JSON (using Jackson).

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/EncoderResultMapper.java
+++ b/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/EncoderResultMapper.java
@@ -1,0 +1,71 @@
+package io.quarkiverse.mcp.server.runtime;
+
+import java.util.List;
+
+import io.quarkiverse.mcp.server.Encoder;
+import io.quarkus.arc.All;
+import io.smallrye.mutiny.Uni;
+
+/**
+ * Result mapper based on {@link Encoder}.
+ *
+ * @param <ENCODED> The type of the value encoded by the encoder
+ * @param <ENCODER> The type of the injected encoder
+ * @param <RESPONSE> The type of the serialized response
+ */
+abstract class EncoderResultMapper<ENCODED, ENCODER extends Encoder<?, ENCODED>, RESPONSE>
+        implements EncoderMapper<Object, RESPONSE> {
+
+    @All
+    List<ENCODER> encoders;
+
+    final EncoderMapper<Uni<Object>, RESPONSE> uni;
+
+    protected EncoderResultMapper() {
+        this.uni = new EncoderMapper<Uni<Object>, RESPONSE>() {
+
+            @Override
+            public Uni<RESPONSE> apply(Uni<Object> uni) {
+                return uni.map(o -> toResponse(EncoderResultMapper.this.convert(o)));
+            }
+        };
+    }
+
+    @Override
+    public Uni<RESPONSE> apply(Object obj) {
+        return Uni.createFrom().item(toResponse(convert(obj)));
+    }
+
+    public EncoderMapper<Uni<Object>, RESPONSE> uni() {
+        return uni;
+    }
+
+    protected abstract RESPONSE toResponse(ENCODED encoded);
+
+    protected ENCODED convert(Object obj) {
+        Class<?> type = obj.getClass();
+        for (ENCODER encoder : encoders) {
+            if (encoder.supports(type)) {
+                ENCODED encoded;
+                try {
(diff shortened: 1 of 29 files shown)


core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/EncoderResultMapper.java
```diff
@@ -0,0 +1,70 @@
...
+import io.quarkiverse.mcp.server.Encoder;
+import io.quarkus.arc.All;
+import io.smallrye.mutiny.Uni;
+
+/**
+ *
+ * @param <ENCODED>
+ * @param <ENCODER>
+ * @param <RESPONSE>
+ */
```
geoand, line 14: This isn't terribly helpful :)

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/ListEncoderResultMapper.java
```diff
@@ -0,0 +1,60 @@
...
+
+import io.quarkiverse.mcp.server.Encoder;
+import io.smallrye.mutiny.Uni;
+
+/**
+ *
+ * @param <ENCODED>
+ * @param <ENCODER>
+ * @param <RESPONSE>
+ */
```
geoand, line 13: Same

docs/modules/ROOT/pages/index.adoc
```diff
@@ -336,7 +342,13 @@ However, the annotated method can also return other types that are converted acc
 * If the method returns a `List` of `Content` implementations or `String`s then the reponse is "success" and contains the list of relevant content objects.
 * The method may return a `Uni` that wraps any of the type mentioned above.
 
-TIP: If a method annotated with `@Tool` throws a `io.quarkiverse.mcp.server.ToolCallException` then the response "failure" and the message of the exception is used as the text of the result content.
+* If it returns `java.lang.String` then the reponse is "success" and contains a single `TextContent`.
+* If it returns an implementation of `Content` then the reponse is "success" and contains a single content object.
+* If it returns a `List` of `Content` implementations or strings then the reponse is "success" and contains a list of relevant content objects.
+* If it returns any other type `X` or `ListX` then `X` is encoded using the `ToolResponseEncoder` and `ContentEncoder` API and afterwards the rules above apply.
```
geoand, line 348:
```suggestion
* If it returns any other type `X` or `List<X>` then `X` is encoded using the `ToolResponseEncoder` and `ContentEncoder` API and afterwards the rules above apply.
```

geoand (commented):
Nice!

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/EncoderResultMapper.java
mkouba: Yep, I should add some description here...

geoand (approved):

PR #93, opened 2025-01-29
stdio: make it possible to disable the transport completely

mkouba:
It's a draft for now - the missing part is disabling the `StdioConfigBuilderCustomizer` that sets `quarkus.log.console.stderr=true`.

CC @maxandersen


mkouba:
Unfortunately, I did not find a way to "disable" the `StdioConfigBuilderCustomizer` at runtime if `quarkus.mcp.server.stdio.enabled=false`. So we just mention this fact in the config reference doc for now.

PR #95, opened 2025-01-30
Code adjustment - typos, method names, links

rsvoboda:
* Rename propmpResultMapper method to promptResultMapper
* Typos fixes
* Fix TrafficLogger to apply textPayloadLimit
* Use System.out.printf instead of String.format and System.out.println combo
* Fix links to the specification
* Update JavaDoc params
* Used method reference instead of lambda


mkouba (approved):
Awesome, thanks!

PR #101, opened 2025-02-07
Introduce WrapBusinessError interceptor binding

### LINKED ISSUE #98: Provide an API to turn any exception thrown in a feature method into a business logic error
Currently, we provide `ToolCallException` that denotes a business logic error and is automatically converted to a failed `ToolResponse` (the message of the exception is used as the text of the result content).

It might be useful to provide an API to transform any exception in a business logic error, i.e. `ToolCallException` for a `@Tool` method.

We could:

- introduce a CDI interceptor; so that a user can annotate a feature method with an interceptor binding such as `@CatchBusinessError` or even `@CatchBusinessError(NullPointerException.class)`
- introduce an `ExceptionMapper` API
- add a config property to turn any exception thrown into a business logic error

I tend to prefer option 1 - a CDI interceptor.

mkouba:
- resolves #98

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/mcp/server/deployment/McpServerProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/mcp/server/deployment/McpServerProcessor.java
@@ -45,6 +45,7 @@ import io.quarkiverse.mcp.server.ResourceResponse;
 import io.quarkiverse.mcp.server.TextContent;
 import io.quarkiverse.mcp.server.TextResourceContents;
 import io.quarkiverse.mcp.server.ToolResponse;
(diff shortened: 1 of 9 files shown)


mkouba:
@maxandersen FYI `@WrapToolBusinessError` can be declared on a method or on a class. And it wraps all exceptions by default. I'm not so sure about the name but did not find anything better.

maxandersen:
How about @WrapOnException ? wether its business or internal is separate concern i feel?

mkouba:
> How about @WrapOnException ? wether its business or internal is separate concern i feel?

Hm, I don't like that. The point is to wrap an exception if it represents a business error. If it doesn't then we don't wrap but log an exception and reply "internal error" to the client.

PR #103, opened 2025-02-10
Programmatic API do add resources, prompts, tools, etc.

### LINKED ISSUE #99: Using programmatically
Hey folks, I'm considering using the Quarkus MCP server for an idea, but I must do so programmatically (i.e.: I cannot use the annotations, as most of the tools, resources, etc need to be defined at runtime). 

Please, do you have any pointers as to how someone could get started with that using the Quarkus MCP server?

mkouba:
- resolves #99


mkouba:
FYI @orpiske ... I will add some docs tomorrow.

orpiske:
That is awesome @mkouba !!!

PR #125, opened 2025-02-25
fix: make proxy a runnable jar from maven

maxandersen:
- make the released proxy jar directly runnable
- remove unneeded --quiet
- document runnable jar coordinates (works when this is released)

### DIFF
--- a/devtools/stdio-sse-proxy/README.md
+++ b/devtools/stdio-sse-proxy/README.md
@@ -57,5 +57,12 @@ java -jar target/quarkus-mcp-stdio-sse-proxy-1.0.0-SNAPSHOT-runner.jar http://my
 You can also run the proxy as a JBang script:
 
 ```shell script
-jbang --quiet src/main/java/io/quarkiverse/mcp/server/proxy/StdioSseProxy.java -h
+jbang src/main/java/io/quarkiverse/mcp/server/proxy/StdioSseProxy.java -h
 ```
+
+or when released to maven use:
+
+```shell script
+jbang io.quarkiverse.mcp:quarkus-mcp-stdio-sse-proxy:RELEASE -h
+```
+
--- a/devtools/stdio-sse-proxy/README.md
+++ b/devtools/stdio-sse-proxy/README.md
@@ -17,16 +17,15 @@ The application can be packaged using:
 This builds a Quarkus uber-jar that can be run directly:
 
 ```shell script
-java -jar target/quarkus-mcp-stdio-sse-proxy-1.0.0-SNAPSHOT-runner.jar --help
+java -jar target/quarkus-mcp-stdio-sse-proxy-1.0.0-SNAPSHOT.jar --help
 ```
 
 Shows the following help message:
 
 ```shell script
-Usage: stdio-sse-proxy [-hV] [--[no-]reconnect] -e=<sseEndpoint> [-s=<sleep>]
-                       [-t=<timeout>]
-  -e, --endpoint=<sseEndpoint>
-                            The URI of the target SSE endpoint
+  Usage: stdio-sse-proxy [-hV] [--[no-]reconnect] [-s=<sleep>] [-t=<timeout>]
+                       <sseEndpoint>
+      <sseEndpoint>         The URI of the target SSE endpoint
   -h, --help                Show this help message and exit.
       --[no-]reconnect      If set to true then the proxy attempts to reconnect
                               if a message endpoint returns http status 400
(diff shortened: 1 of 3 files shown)


devtools/stdio-sse-proxy/src/main/java/io/quarkiverse/mcp/server/proxy/StdioSseProxy.java
```diff
@@ -1,7 +1,7 @@
 //usr/bin/env jbang "$0" "$@" ; exit $?
 //DEPS io.quarkus.platform:quarkus-bom:3.18.3@pom
 //DEPS io.quarkus:quarkus-picocli
-//Q:CONFIG quarkus.log.console.stderr=true
+//FILES application.properties=../../../../../../resources/application.properties
```
mkouba, line 4: Why this change? I don't like the `../../../../../../resources/application.properties` part.

devtools/stdio-sse-proxy/README.md
```diff
@@ -51,5 +51,12 @@ java -jar target/quarkus-mcp-stdio-sse-proxy-1.0.0-SNAPSHOT-runner.jar -e http:/
...
-jbang --quiet src/main/java/io/quarkiverse/mcp/server/proxy/StdioSseProxy.java -h
-```
\ No newline at end of file
+jbang src/main/java/io/quarkiverse/mcp/server/proxy/StdioSseProxy.java -h
+```
+
+or when released to maven use:
+
+```shell script
+jbang io.quarkiverse.mcp:quarkus-mcp-stdio-sse-proxy:RELEASE -h
```
mkouba, line 60: What does the `:RELEASE` stand for? Is it handled specifically by JBang?

devtools/stdio-sse-proxy/src/main/java/io/quarkiverse/mcp/server/proxy/StdioSseProxy.java
maxandersen: so it uses the same application.properties as the mvn build.

but ultimately if just getting this published to maven central it does not need to be a jbang script so can remove it all

mkouba: Hm, `quarkus.log.console.stderr=true` is the only property that matters here and I'd like to keep the possibility to run this as a jbang script.

devtools/stdio-sse-proxy/README.md
mkouba: Is it? I've never heard about it. I mean, `io.quarkiverse.mcp:quarkus-mcp-stdio-sse-proxy:RELEASE` is not a valid Maven artifact coordinate (at least according to chatgpt and https://maven.apache.org/guides/mini/guide-naming-conventions.html). `release` is a valid qualifier but nothing more. But maybe I'm missing something?

devtools/stdio-sse-proxy/src/main/java/io/quarkiverse/mcp/server/proxy/StdioSseProxy.java
mkouba: > sure - but above keeps letting you run it as a jbang script.

Yes, but that's the intention. You can run the deployed artifact or the latest version as a script.

> anyhow - no big deal; I can remove it if you want/need?

+1

maxandersen: ok so just tested this and this need to use //FILES to work because https://github.com/quarkusio/quarkus/issues/44466 is still an issue.

runtime/buildtime config does not get honored since the config system changed.

mkouba: Funny, it worked just fine when I tested the PR but maybe I missed something... :thinking:

mkouba: If I understand the description of https://github.com/quarkusio/quarkus/issues/44466 correctly then `//Q:CONFIG` works but the solution is suboptimal and should be refactored.

@maxandersen I will remove this change and update the readme.

mkouba (approved):

devtools/stdio-sse-proxy/src/main/java/io/quarkiverse/mcp/server/proxy/StdioSseProxy.java
maxandersen: well it means that //Q:CONFIG for many cases will only take effect if jbang does a build - not when just running jar directly. Basically - //Q:CONFIG is not useful unless doing development.

devtools/stdio-sse-proxy/README.md
maxandersen: That is for *plugins* it was explicitly removed as makes builds irreproducible. still possible for resolver because it is useful - like in cases like this. 

its like `docker pull xyz:latest` or `sdk install quarkus` where you just get the latest.

PR #129, opened 2025-02-26
feat: have proxy default use quarkus-dev coordinates and otherwise have first argument be uri

maxandersen:
makes it easier to run the proxy as it will default to quarkus devmode coordinates and otherwise have as first argument.

all other flags stay as is.


mkouba (approved):
Nice!

PR #138, opened 2025-02-28
Support Optional parameters

geoand:


core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/ToolManagerImpl.java
```diff
@@ -108,6 +108,11 @@ Object generateSchema(Type type, String description) {
             if (description != null && !description.isBlank()) {
                 objectNode.put("description", description);
             }
+            if (Types.isOptional(type)) {
+                objectNode.remove("type");
+                objectNode.putIfAbsent("type",
+                        objectNode.withObjectProperty("properties").withObjectProperty("value").get("type"));
+            }
```
mkouba, line 115: Hm, the generated schema for `MyTools#alpha()` looks like:

```json
{
      "name" : "alpha",
      "description" : "",
      "inputSchema" : {
        "type" : "object",
        "properties" : {
          "price" : {
            "properties" : {
              "value" : {
                "type" : "integer"
              }
            },
            "description" : "Define the price...",
            "type" : "integer"
          }
        },
        "required" : [ "price" ]
      }
    }
```
Which does not look correct. I would expect something like:

```json
    {
      "name" : "alpha",
      "description" : "",
      "inputSchema" : {
        "type" : "object",
        "properties" : {
          "price" : {
            "description" : "Define the price...",
            "type" : "integer"
          }
        },
        "required" : [ ]
      }
    }
```

`properties` would only make sense for `object` type.
 
Note that the price is *required*. I think that for `Optional` it should only be required if the `@ToolArg` specifies this directly. WDYT?

mkouba (commented):
This is pretty cool! I've added a comment for json schema generation...

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/ToolManagerImpl.java
mkouba: This should be addressed in the second commit.

geoand: Indeed, thanks!

PR #151, opened 2025-03-14
docs: clarify that MCP client is provided by the LangChain4j project

mkouba:


README.md
```diff
@@ -2,13 +2,15 @@
...
 <!-- ALL-CONTRIBUTORS-BADGE:END -->
-
 [![Version](https://img.shields.io/maven-central/v/io.quarkiverse.mcp/quarkus-mcp-server-parent?logo=apache-maven&style=flat-square)](https://central.sonatype.com/artifact/io.quarkiverse.mcp/quarkus-mcp-server-parent)
 
 _"[Model Context Protocol](https://modelcontextprotocol.io/) (MCP) is an open protocol that enables seamless integration between LLM applications and external data sources and tools."_
 
 This extension provides declarative and programmatic APIs that enable developers to implement the MCP server features easily.
 
+> [!NOTE]  
+> The [LangChain4j](https://github.com/langchain4j/langchain4j) project provides a programmatic API for MCP clients.
```
jmartisk, line 12: Given that in many cases you don't need any Java code at all to get MCP working, I guess this could be worded differently. Maybe that it provides a Java implementation of MCP client functionality and its integration into AI-infused applications...?

mkouba: Hm, but that's a functionality provided by the `quarkus-langchain4j`, or? I'd like to keep this sentence about programmatic API because "integration into AI-infused applications" is IMO a higher-level concept. 

How about something like:

_The LangChain4j project provides a programmatic API for MCP clients. The `quarkus-langchain4j` extension provides the MCP client functionality and its integration into AI-infused applications._

jmartisk: I would not call it programmatic API, the amount of code  needed to use it is minimal if any at all (whether you are using langchain4j directly or the quarkus extension). 
I also wouldn't mention any difference between vanilla langchain4j and the quarkus extension apart from the fact that the quarkus extension is the preferred way to use it in Quarkus apps :)

mkouba: > I would not call it programmatic API, the amount of code needed to use it is minimal if any at all (whether you are using langchain4j directly or the quarkus extension).

Well, this is what I call a programmatic API:

```java
McpTransport transport = new HttpMcpTransport.Builder()
    .sseUrl("http://localhost:8080/mcp/sse")
    .build();
McpClient mcpClient = new DefaultMcpClient.Builder()
    .transport(transport)
    .build();
ToolProvider toolProvider = McpToolProvider.builder().mcpClients(List.of(mcpClient)).build();
ToolProviderResult res = toolProvider.provideTools(null);
```

I know that it's a low-level API and it's not needed if you write an AI-infused app but it's available and may be used to integrate/test MCP outside langchain4j.

> I also wouldn't mention any difference between vanilla langchain4j and the quarkus extension apart from the fact that the quarkus extension is the preferred way to use it in Quarkus apps :)

Ok, didn't know it works in the same way in vanilla langchain4j.

How about this:

_The LangChain4j project provides the MCP client functionality, either as a low-level programmatic API or as a full-fledged integration into AI-infused applications._

jmartisk: I don't call that piece of code programmatic API, it's just a few lines of scaffolding to plug a MCP client into an application (and Quarkus saves you even from these few lines, because it generates a synthetic `ToolProvider`). After that, actually _using_ the client doesn't require any code, that is what I mean. The `toolProvider.provideTools` call is technically public API, but in most cases it shouldn't be necessary to use, the AI service will do it for you.

jmartisk: But ok, if you don't want to hide the fact that you can use it manually (`toolProvider.provideTools` etc), then I guess your current wording is fine

mkouba: > I don't call that piece of code programmatic API, it's just a few lines of scaffolding to plug a MCP client into an application (and Quarkus saves you even from these few lines, because it generates a synthetic `ToolProvider`). After that, actually _using_ the client doesn't require any code, that is what I mean. The `toolProvider.provideTools` call is technically public API, but in most cases it shouldn't be necessary to use, the AI service will do it for you.

I know what you mean...

> But ok, if you don't want to hide the fact that you can use it manually (`toolProvider.provideTools` etc), then I guess your current wording is fine

... and this is exactly what I mean ;-).

Issue #155, opened 2025-03-24
Make it easier to start the mcp-server-stdio from the CLI or a CLI plugin


iocanel:
I tried to use the `mcp-server-stdio` from within the quarkus-backstage CLI. It worked great, but it completly broke all other commands that were part of the CLI as it was hijacking `System.out`. The only workaround I found was to use `mcp-sever-core` and copy a couple of classes: 
https://github.com/quarkiverse/quarkus-backstage/tree/main/cli/src/main/java/io/quarkiverse/backstage/cli/mcp

PR #156, opened 2025-03-24
feat: Allow mcp-server-stdio to be explicilty started.

iocanel:
Resolves: #155 

The pull request introduces a build time config parameter that controls the initialization of the `mcp-server-stdio` this allows users to have more control over when the server is started.

For my use case, this works like a charm.

### DIFF
--- a/transports/stdio/runtime/src/main/java/io/quarkiverse/mcp/server/stdio/runtime/StdioMcpMessageHandler.java
+++ b/transports/stdio/runtime/src/main/java/io/quarkiverse/mcp/server/stdio/runtime/StdioMcpMessageHandler.java
@@ -9,6 +9,7 @@ import java.util.Base64;
 import java.util.UUID;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import jakarta.inject.Singleton;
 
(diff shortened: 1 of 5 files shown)


transports/stdio/runtime/src/main/java/io/quarkiverse/mcp/server/stdio/runtime/StdioMcpMessageHandler.java
```diff
@@ -57,7 +57,7 @@ protected StdioMcpMessageHandler(McpRuntimeConfig config, ConnectionManager conn
         this.vertx = vertx;
     }
 
-    void initialize(PrintStream stdout, McpRuntimeConfig config) {
+    public void initialize(PrintStream stdout, McpRuntimeConfig config) {
```
mkouba, line 60: Since we're making this public and the `StdioMcpMessageHandler` is a CDI bean, could you pls add the `private final AtomicBoolean initialized = new AtomicBoolean()` field and only initialize if `initialize.compareAndSet(false, true)` returns `true`?

iocanel: Done.

mkouba: Cool, thanks!

mkouba (approved):

PR #177, opened 2025-04-03
stdio: fix list changed notifications

mkouba:


transports/stdio/runtime/src/main/java/io/quarkiverse/mcp/server/stdio/runtime/StdioMcpMessageHandler.java
```diff
@@ -65,6 +65,7 @@ public void initialize(PrintStream stdout, McpRuntimeConfig config) {
             String connectionId = Base64.getUrlEncoder().encodeToString(UUID.randomUUID().toString().getBytes());
             StdioMcpConnection connection = new StdioMcpConnection(connectionId, config.clientLogging().defaultLevel(),
                     trafficLogger, config.autoPingInterval(), stdout, vertx);
+            connectionManager.add(connection);
```
mkouba, line 68: @jmartisk this is the actual fix :facepalm:

jmartisk (approved):
Thanks :)

PR #199, opened 2025-04-09
Add secure-http-server demo

sberyozkin:
Show to get an access token in devmode and use it with curl to access the SSE endpoint


sberyozkin:
@maxandersen @mkouba This demo should be ready to go quite soon, if it were possible to use the access token from MCP server DevUI then it would be nice

sberyozkin:
@mkouba Can you please merge this one too, so that I can link to his example, and we can update the link later if it gets moved to the mcp-servers repo

PR #212, opened 2025-04-14
Initial Roots support

mkouba:


maxandersen:
Cool. Is that mcpconnection available on tools calls too?

Having some of this available on a request or session starts being useful :)

mkouba:
> Cool. Is that mcpconnection available on tools calls too?

Yes, both the `McpConnection` and `Roots` can be injected into params of a `@Tool` method (or any feature method)...

> Having some of this available on a request or session starts being useful :)

Yes, it's still on my radar. However, it does not really fit the "stateless" scenario of the new streamable HTTP transport :shrug:. Also it could lead to the "typical" context propagation issues ;-).

PR #220, opened 2025-04-17
devui: add input field for bearerToken header

mkouba:
- also make it possible to force a new MCP session

### DIFF
--- a/sse-client/src/main/java/io/quarkiverse/mcp/server/sse/client/SseClient.java
+++ b/sse-client/src/main/java/io/quarkiverse/mcp/server/sse/client/SseClient.java
@@ -1,6 +1,7 @@
 package io.quarkiverse.mcp.server.sse.client;
 
 import java.io.EOFException;
+import java.io.IOException;
 import java.net.ConnectException;
 import java.net.URI;
 import java.net.http.HttpClient;
@@ -109,9 +110,17 @@ public abstract class SseClient {
         @Override
         public void onError(Throwable t) {
             Throwable root = getRootCause(t);
-            if (!(root instanceof EOFException)) {
-                LOG.errorf(t, "Error in SSE stream");
+            if (root instanceof IOException io) {
+                if (io.getMessage().contains("Request cancelled")) {
+                    // This is ok - no need to log anything
+                    return;
+                }
             }
+            if (root instanceof EOFException) {
+                // EOF is not a problem
+                return;
+            }
+            LOG.errorf(t, "Error in SSE stream");
         }
 
         @Override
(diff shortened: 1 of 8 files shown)


sberyozkin:
Thanks @mkouba, I tested it and so far could not manage to get it working.

The easiest way to reproduce is with #199 (and the `999-SNAPSHOT` MCP SSE transport dependency).
First, follow README to confirm it works with `curl`.

The only thing that should be done differently (I'll have to fix README) is, in the step `Login and Copy access token`, you should do it with Dev UI: 

https://quarkus.io/guides/security-openid-connect-dev-services#dev-services-for-keycloak

Click on `Keycloak provider` in the OIDC card, then on `Login into Single Page Application`, then enter `alice` and `alice` and copy the access token from the DevUI SPA page.

And once you confirm it works with `curl`, re-try with MCP Server DevUI,

I was wondering if MCP Server DevUI was working with a devmode specific endpoint, different from `/mcp/sse` ? 
The other thing, would it make sense to offer a `Bearer token` option earlier, as I'm assuming, if the MCP server security will be required then not only tools but also prompts, resources, would also be expected to be protected.  We can definitely do it later though if proves necessary.

Thanks

mkouba:
> Thanks @mkouba, I tested it and so far could not manage to get it working.

It should work now (I tested this PR with yours ;-) - we did not use the bearer token for the initial `mcp/sse` request. However, if it's not your first request then it only works if you check the "Force new session" option.

> The easiest way to reproduce is with #199 (and the `999-SNAPSHOT` MCP SSE transport dependency). First, follow README to confirm it works with `curl`.
> 
> The only thing that should be done differently (I'll have to fix README) is, in the step `Login and Copy access token`, you should do it with Dev UI:
> 
> https://quarkus.io/guides/security-openid-connect-dev-services#dev-services-for-keycloak
> 
> Click on `Keycloak provider` in the OIDC card, then on `Login into Single Page Application`, then enter `alice` and `alice` and copy the access token from the DevUI SPA page.
> 
> And once you confirm it works with `curl`, re-try with MCP Server DevUI.
> 
> I was wondering if MCP Server DevUI was working with a devmode specific endpoint, different from `/mcp/sse` ? 

> The other thing, would it make sense to offer a `Bearer token` option earlier, as I'm assuming, if the MCP server security will be required then not only tools but also prompts, resources, would also be expected to be protected. We can definitely do it later though if proves necessary.

Yes, the current UI is a bit clumsy but I'd like to keep the possibility to be able to create a new MCP session. In theory, we could just keep the last bearer token used or something like that...

sberyozkin:
Hey @mkouba Perfect, it works and I've done screenshots for the blog post. Please merge

sberyozkin (approved):

Issue #217, opened 2025-04-16
[docs] Unclear Resources Example


Mr-Steel:
Hi,
I am following the examples and stumbled across this section https://docs.quarkiverse.io/quarkus-mcp-server/dev/#_resources
The example, like the one in the readme, refers to some kind of variable or constant called "Paths.ALPHA"

<code>
    @Resource(uri = "file:///project/alpha") 
    BlobResourceContents alpha() {
        return BlobResourceContents.create("file:///project/alpha", Files.readAllBytes(Paths.ALPHA));
    }
</code>

It is repeated in the Resource Template section. Might be a copy and paste error from an earlier version.

mkouba:
It's a dummy constant that holds some arbitrary `Path`. We could replace it with something like `Path.of("alpha.txt")`...

Mr-Steel:
Ah, I see - yeah I think this is much clearer to the reader, Thank you!

mkouba:
> Ah, I see - yeah I think this is much clearer to the reader, Thank you!

Would you care to send a pull request? ;-)

PR #225, opened 2025-04-23
#217:[docs] Unclear Resources Example

Mr-Steel:
#217: improved the documentation a bit with the suggestion from @mkouba -> instead of using what looks like a constant (Paths.ALPHA) using a generic Path reference


mkouba (approved):

mkouba:
@Mr-Steel merged, thanks!

PR #231, opened 2025-04-26
Add parallel build steps for sample projects in CI workflow

kpavlov:
Make sure that samples are buildable.

See https://github.com/kpavlov/quarkus-mcp-server/pull/1 for build execution.

This pull request introduces several updates to enhance the build process, improve compatibility, and standardize configurations across the project. The key changes include adding parallelized sample project builds, updating Java and Maven configurations, and introducing a `Makefile` for streamlined build commands.

### Build Process Enhancements:
* Added a new step in `.github/workflows/build.yml` to build sample projects in parallel using a script that verifies each sample's build status. This ensures that all sample projects are validated during the CI process.
* Introduced a `Makefile` with a `build` target to consolidate and simplify the build process for both Java and native builds, as well as sample project builds.

### Java and Maven Configuration Updates:
* Updated the `.java-version` file to specify Java 17, aligning the project with the latest supported LTS version.
* Added JVM options in `.mvn/jvm.config` to improve compatibility and performance, including memory settings and module access configurations.
* Standardized JVM and Maven configurations for sample projects (`secure-mcp-sse-server` and `weather`), including memory settings, module access, and parallel build options. [[1]](diffhunk://#diff-1543c3387c9f94d213a35133664c0c9ddef12b459173507c6f761cd666bd95e7R1-R5) [[2]](diffhunk://#diff-5eb5f7479ec7ca23d9f0f6d3e085aa3f74145026849c9ecaaf4f09db74d1ec87R1)

### Dependency and Compiler Updates:
* Updated the Maven Compiler Plugin version in the `pom.xml` files of sample projects (`secure-mcp-sse-server` and `weather`) to `3.14.0` and set the Java compiler release to 17 for compatibility with the updated Java version. [[1]](diffhunk://#diff-527a45d03f620951fbbc6d281ece6d97c89f1c7c110accac4072b6b367e22ea8L9-R10) [[2]](diffhunk://#diff-e0e482771fd2c100c4d86a6ba51d3c6f5aba54ebefed1452ff3580de5b2f3d3eL9-R9)

PR #233, opened 2025-04-26
Upgrade Maven compiler plugin and adjust Java release version

kpavlov:
Updated the compiler plugin to version 3.14.0 across projects for improved compatibility and features. Adjusted the Java release version in the secure-mcp-sse-server module from 21 to 17 to align with supported configurations.


kpavlov:
Extracted from #231

PR #256, opened 2025-05-14
core: fail the build if a feature annotation is used incorrectly

mkouba:
- i.e. when a feature annotations is declared on a static method, or a method declared on an interface


geoand (approved):
Makes sense!

PR #257, opened 2025-05-14
core: determine set of visible/accesible features

### LINKED ISSUE #248: Filter MCP tools, resources and prompts before listing
When an MCP client issues a resources/tools/prompt list request, I would like to be able to be able to filter the list passed to the MCP client. 

For instance, consider this scenario: an application that is using the programmatic API and is exposing 3 tools (`listOrder`, `readOrder`, `createOrder`). 

When a new MCP connection is made, I don't want to list `createOrder` unless the user/agent requesting the listing is authorized to use that tool. 

Consider as well that the rules for filtering this out may be related to other things other than the authorization (i.e.: not exposing certain tools if the origin of the connection is a sanctioned country). 

As such, I'd like to be able to do something like this:

```
interface McpFilter {
     List<McpResource> filter(McpRequest mcpRequest, List<McpResource> input)
},

toolManager.setFilter(new McpFilter {
    ....
});
```
Then, when the `list` request is issued by the MCP client, the toolManager would provide the values returned by the `McpFilter` instance.

mkouba:
- i.e. to filter tools, resources, prompts and resource templates
- resolves #248

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/mcp/server/deployment/McpServerProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/mcp/server/deployment/McpServerProcessor.java
@@ -51,15 +51,19 @@ import io.quarkiverse.mcp.server.EmbeddedResource;
 import io.quarkiverse.mcp.server.ImageContent;
 import io.quarkiverse.mcp.server.ModelHint;
 import io.quarkiverse.mcp.server.ModelPreferences;
+import io.quarkiverse.mcp.server.PromptFilter;
 import io.quarkiverse.mcp.server.PromptMessage;
 import io.quarkiverse.mcp.server.PromptResponse;
 import io.quarkiverse.mcp.server.ResourceContents;
+import io.quarkiverse.mcp.server.ResourceFilter;
 import io.quarkiverse.mcp.server.ResourceResponse;
+import io.quarkiverse.mcp.server.ResourceTemplateFilter;
 import io.quarkiverse.mcp.server.Role;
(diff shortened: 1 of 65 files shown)


mkouba:
Docs are still missing.

mkouba:
Docs added. 

@orpiske Let me know if this API suits your needs.

orpiske:
Hello @mkouba, thank you for considering the change! It is on top of my list to take a look at this today. I'll take a look and let you know no later than tomorrow, OK?

mkouba:
> Hello @mkouba, thank you for considering the change! It is on top of my list to take a look at this today. I'll take a look and let you know no later than tomorrow, OK?

That would be great :+1:

orpiske:
@mkouba I've just tested it and it looks really cool! 

I tested it here with this filter:

```
public class TestFilter implements ToolFilter {
    private static final Logger LOG = Logger.getLogger(TestFilter.class);

    @Inject
    HttpServerRequest request;

    @Override
    public boolean test(ToolManager.ToolInfo tool, McpConnection connection) {
        LOG.infof("Calling/Listing tool %s", tool.name());
        if ("jira".equals(tool.name())) {
            return false;
        }

        final String header = request.getHeader("Content-type");
        return true;
    }
}
```

It seems to me that I have all the information I need in order to filter the tools according to the request I receive. 

Thanks for this!

orpiske (approved):

mkouba:
:+1: Cool, we could always improve/enhance the API later...

PR #285, opened 2025-06-03
Support multiple MCP server configurations in one application

### LINKED ISSUE #262: Support multiple MCP server configurations in one application
Recently, I closed an [issue](https://github.com/quarkiverse/quarkus-mcp-server/issues/239) about supporting multiple MCP server configurations as "not planned". Today, @sberyozkin asked a similar [question on zulip](https://quarkusio.zulipchat.com/#narrow/channel/187038-dev/topic/MCP/near/519362591). Therefore, I've created this tracking issue to collect requirements, ideas and implementation considerations.

### API

We would need a declarative way to bind a feature to a specific config, e.g. `@McpServer("foo")` and a also make it possible to set the config name when registering a feature programmatically. There would also be a default config, used when no explicit config is defined. 

The question is - do we bind a feature to a single MCP server config only? I think that it would make sense but it might not be practical for the use case described by @sberyozkin (multiple entry points and security).

### Transports

It's clear that multiple configs would only make sense for transports that support multiple MCP clients (e.g. HTTP/SSE). It's completely useless for the `stdio` transport.

mkouba:
- resolves #262


mkouba:
@sberyozkin FYI ;-)

sberyozkin:
Thanks @mkouba  Let me have a look and test with 2 OIDC providers, will try tomorrow

sberyozkin:
Hi Martin, I think this should work perfectly for the took specific security, given something like

```
quarkus.mcp.server.sse.root-path=/alpha/mcp <3>
quarkus.mcp.server.bravo.sse.root-path=/bravo/mcp <4>
```

Let's say it is `bravo`, then the root endpoint is `/bravo/mcp` and the messages endpoint becomes `/bravo/mcp/messages/*` ?

How will the root and message endpoints for `bravo` be updated if it is Streamable HTTP ?

I'd like to do a quick test asap, just trying to get closer to completing one of Quarkus issues.

mkouba:
> Let's say it is `bravo`, then the root endpoint is `/bravo/mcp` and the messages endpoint becomes `/bravo/mcp/messages/*`, when the SSE transport is used ?

Yes, exactly. And the SSE endpoint is `/bravo/mcp/sse`.



> How will the root and message endpoints for `bravo` be updated if it is Streamable HTTP ?

For the streamable HTTP the MCP endpoint is `/bravo/mcp`. 

> I'd like to do a quick test asap, just trying to get closer to completing one of Quarkus issues.

I will merge this PR now. We could implement improvements once you have the test app.

sberyozkin:
Sounds good, thanks @mkouba

PR #288, opened 2025-06-09
sse: make it possible to include query params in the message endpoint

### LINKED ISSUE #287: Enhancement. In sse pass query parmater to mesages route
Hi when we call the sse endpoint https://localhost:8443/mcp/sse?param=param1&test=test1 they are lost in the following call to messages route , it could be useful to keep them in the redirect url so they can be accessed during tool execution and other places in the code. 

Do you think it's feasible? The change should be something like :
`   String endpointPath = mcpPath.endsWith("/") ? (mcpPath + "messages/" + id) : (mcpPath + "/messages/" + id);
                LOG.debugf("POST endpoint path: %s", endpointPath);
                MultiMap params = ctx.request().params();
				if (params != null && !params.isEmpty()) {
					StringBuilder sb = new StringBuilder();
					params.forEach(entry -> {
						if (sb.length() > 0) {
							sb.append("&");
						}
						sb.append(entry.getKey()).append("=").append(entry.getValue());
					});
					endpointPath += "?" + sb.toString();
				}
                connection.sendEvent("endpoint", endpointPath);
`

mkouba:
- with `quarkus.mcp.server.sse.message-endpoint.include-query-params=true`
- resolves #287

### DIFF
--- a/docs/modules/ROOT/pages/includes/quarkus-mcp-server-sse.adoc
+++ b/docs/modules/ROOT/pages/includes/quarkus-mcp-server-sse.adoc
@@ -36,5 +36,31 @@ endif::add-copy-button-to-env-var[]
 |string
 |`/mcp`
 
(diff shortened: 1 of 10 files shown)


mkouba:
@psini FYI the feature is disabled by default.

mkouba:
@sberyozkin @cescoffier I wonder if this feature, i.e. copy the query params from the initial HTTP request to an SSE endpoint to the relevant message endpoint, could be considered a security vulnerability in some special use cases?

Issue #307, opened 2025-06-23
Implement cancellation of in-progress requests


mkouba:
See also the spec: https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/cancellation

mkouba:
> I'm interested in taking up, is that fine to pick up?

Contributions are welcome! So yes, it's fine.

> If yes any suggestions on starting point / some insights will be really great.

I think that we'll need a new `@Singleton` CDI bean, maybe named `Cancellations` with a single public method `boolean isCancelled(RequestId requestId)` and this method should return `true` if a client sends a cancellation request for this request id. So the bean will need to store cancellation requests for in-progress requests for each client, i.e. basically a map of connection id -> list of cancellation requests. And the map should be updated whenever a request is processed or the client/connection is closed.

This bean could be injected directly in the bean class that declares features, or as a method parameter.

And the author of the feature would need to check `Cancellations#isCancelled()` periodically.

WDYM?

PR #319, opened 2025-06-28
introduce Cancellations for request cancellations

prakashelango:
#307


mkouba (commented):
This is a good start.

I think that we should:
1. Extract the `Cancellations` interface in the `io.quarkiverse.mcp.server` package.
2. Rename this bean to `CancellationsImpl` and make it package-private and implement `Cancellations` (note that `io.quarkiverse.mcp.server.runtime` is not considered public API)
3. Use the `io.quarkiverse.mcp.server.RequestId` in this API
4. Add some tests ;-)

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/Cancellations.java
```diff
@@ -0,0 +1,62 @@
...
+     */
+    public void cancelRequest(String connectionId, Object requestId) {
+        cancellationsByConnection.computeIfAbsent(connectionId, k -> ConcurrentHashMap.newKeySet())
+                .add(requestId);
+    }
+
+    /**
+     * Remove a request from tracking (when request completes)
+     */
+    public void removeRequest(String connectionId, Object requestId) {
```
mkouba, line 46: Also this method should not be part of the public API; it should be called by the internal code.

prakashelango:
@mkouba Thanks for the review, Have addressed comments PTAL.

core/runtime/src/main/java/io/quarkiverse/mcp/server/Cancellations.java
```diff
@@ -0,0 +1,24 @@
...
+    boolean isCancelled(RequestId requestId);
+
+    /**
+     * Check if a request has been cancelled for a specific connection
+     *
+     * @param connectionId the connection identifier
+     * @param requestId the request identifier
+     * @return true if the request has been cancelled, false otherwise
+     */
+    boolean isCancelled(String connectionId, RequestId requestId);
```
mkouba, line 23: While looking at this API, I wonder if from user POV it wouldn't we easier to consume an interface like:

```java
interface Cancellation {
   boolean isRequested();
}
```

direclty in a feature method, i.e. something like:

```java
class Tools {

   String foo(Cancellation cancellation) {
      // do something
      if (cancellation.isRequested()) {
         throw new OperationCancelledException();
      }
   }

}
```
This way a user would not need to pass neither the request id nor the connection id :thinking:.

mkouba:
@prakashelango I've created an alternative PR with the API proposal I mentioned in https://github.com/quarkiverse/quarkus-mcp-server/pull/319#discussion_r2204980785. Let me know what you think about this. Thanks!

core/runtime/src/main/java/io/quarkiverse/mcp/server/Cancellations.java
mkouba: Ok, it would be great if you could take a look at https://github.com/quarkiverse/quarkus-mcp-server/pull/332.

prakashelango:
Took at look at the pr it's more aligned. can we close this one?

PR #473, opened 2025-10-29
Warn if CORS is disabled for Streamable HTTP, update security docs

### LINKED ISSUE #452: Server should validate the origin header
As per https://modelcontextprotocol.io/specification/draft/basic/transports#security-warning:

> Servers MUST validate the Origin header on all incoming connections to prevent DNS rebinding attacks

Proposal would be to check the header (if it's present) against the server domain (and maybe an additional list of domains passed as property?).

If this is on interest, I can probably work on something (not before next week though...).

sberyozkin:
Fixes #452.

This PR warns if no CORS is disabled in either prod or dev mode, to minimize the noise during test runs.
The security docs are also updated

### DIFF
--- a/transports/sse/runtime/src/main/java/io/quarkiverse/mcp/server/sse/runtime/StreamableHttpMcpMessageHandler.java
+++ b/transports/sse/runtime/src/main/java/io/quarkiverse/mcp/server/sse/runtime/StreamableHttpMcpMessageHandler.java
@@ -14,6 +14,8 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import jakarta.enterprise.inject.Instance;
 import jakarta.inject.Singleton;
 
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.ConfigProvider;
 import org.jboss.logging.Logger;
 
 import io.quarkiverse.mcp.server.CompletionManager;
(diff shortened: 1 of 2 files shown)


transports/sse/runtime/src/main/java/io/quarkiverse/mcp/server/sse/runtime/StreamableHttpMcpMessageHandler.java
```diff
@@ -77,6 +79,7 @@ public class StreamableHttpMcpMessageHandler extends McpMessageHandler<HttpMcpRe
     private final CurrentIdentityAssociation currentIdentityAssociation;
 
     StreamableHttpMcpMessageHandler(McpServersRuntimeConfig config,
+            VertxHttpConfig httpConfig,
```
mkouba: I'm not sure it's reasonable to depend on the `VertxHttpConfig`. There's no clear opinion whether config mappings are part of the public API, and in this case the package  `io.quarkus.vertx.http.runtime` is IMO not a public API. Recently we tried to avoid this kind of dependencies (e.g. using the `HttpRootPathBuildItem`).

I think that something like `ConfigProvider.getConfig().getOptionalValue("quarkus.http.cors.enabled", Boolean.class)` might be more robust. WDYT?

sberyozkin: HI Martin, done, I can update it further if you'd like, np at all

mkouba (approved):

PierreBtz:
Thanks!!!

PR #494, opened 2025-11-14
core: fix resource template completion

### LINKED ISSUE #493: Resource Template Completion not working if the client does not pass the template name in the reference
When playing around with Resource Template Completion using MCP Inspector, I noticed that the resource template reference it sends in the completion request does not include the template name:

```JSON
{
  "method": "completion/complete",
  "params": {
    "argument": {
      "name": "id",
      "value": "1"
    },
    "ref": {
      "type": "ref/resource",
      "uri": "location://{id}"
    },
    "context": {
      "arguments": {}
    }
  }
}
```

As a matter of fact, this behavior is in conformance to the [specs](https://modelcontextprotocol.io/specification/2025-06-18/server/utilities/completion#reference-types) - while prompts are referenced by name, resource templates are referenced by `uri`. But it looks like the `io.quarkiverse.mcp.server.runtime.CompletionMessageHandler` (also, the `io.quarkiverse.mcp.server.test.complete.ResourceTemplateCompleteTest`) assume the `name` field is present and resolve it to `null`. As a result, the declared completion has no way to be found by the key in `io.quarkiverse.mcp.server.runtime.ResourceTemplateCompletionManagerImpl#completions` map, and the server fails with:

```JSON
{
  "error": "MCP error -32602: Resource template completion does not exist: null_id"
}
```

I'm pretty insecure about this conclusion, but it looks to me like this feature might not work as expected by the specs. I might take a stab at fixing it, but I'd rather someone confirmed this observation.

mkouba:
- fixes #493

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/mcp/server/deployment/McpServerProcessor.java
+++ b/core/deployment/src/main/java/io/quarkiverse/mcp/server/deployment/McpServerProcessor.java
@@ -446,6 +446,25 @@ class McpServerProcessor {
             }
         }
 
+        // Check duplicate uris for resource templates
+        List<FeatureMethodBuildItem> resourceTemplates = found.get(RESOURCE_TEMPLATE);
+        if (resourceTemplates != null) {
+            Map<String, List<FeatureMethodBuildItem>> byUri = resourceTemplates.stream()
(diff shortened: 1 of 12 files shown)


mkouba:
CC @tknyziak

tknyziak:
Impressive speed 💪 I can confirm that my custom server based on the version built from your pull request works as expected with the same MCP Inspector setup I've used before. Good work!

PR #519, opened 2025-12-08
streamable: make it possible to perform dummy initialization

mkouba:
- and relax the requirement for client initialization
- can be used to simulate stateless communication

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/McpMessageHandler.java
+++ b/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/McpMessageHandler.java
@@ -196,18 +196,28 @@ public abstract class McpMessageHandler<MCP_REQUEST extends McpRequest> {
         String method = message.getString("method");
         JsonObject params = message.getJsonObject("params");
 
-        // The first message must be "initialize"
-        // However, in the dev mode if an MCP client attempts to reconnect an SSE connection but does not reinitialize propertly,
-        // we could perform a "dummy" initialization
         if (!INITIALIZE.equals(method)) {
-            if (LaunchMode.current() == LaunchMode.DEVELOPMENT && serverConfig(mcpRequest).devMode().dummyInit()) {
-                InitialRequest dummy = new InitialRequest(new Implementation("dummy", "1", null),
+            // Normally the first message must be "initialize"
+
+            // However, we could create a synthetic initial request and perform a dummy initialization
+            InitialRequest dummy = null;
+            if (LaunchMode.current() == LaunchMode.DEVELOPMENT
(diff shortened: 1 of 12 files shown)


mkouba:
@jmartisk @cescoffier WDYT about this workaround? 

The stateless aspect of the HTTP transport is not part of the spec yet but I assume that we'll have to do something similar, except that instead of the dummy init we'll have to create the initial request based on some HTTP headers (client capabilities, etc.). There is already the `MCP-Protocol-Version` header which we don't support yet.

What I haven't considered much yet is which functionalities are likely to fail...

jmartisk:
So IIUC, this will assume that the client "initializes" declaring no capabilities? (maybe the docs should be more concrete about this?) 
AFAIK, the langchain4j client always re-initializes (at least I hope so), do you know of any clients that don't do it? (asking because if clients don't actually do this, then it makes no sense to apply this hack)

mkouba:
> So IIUC, this will assume that the client "initializes" declaring no capabilities? (maybe the docs should be more concrete about this?)

Yes, no client capabilities. The docs for config property state  _"The capability negotiation and protocol version agreement is completely skipped."_ but you're right, we could make it more explicit.

> AFAIK, the langchain4j client always re-initializes (at least I hope so), do you know of any clients that don't do it? (asking because if clients don't actually do this, then it makes no sense to apply this hack)

This "hack" is intended for clients that expect stateless communication, i.e. send a `tools/call` message without initialization (no `Mcp-Session-Id` header etc.). Although it does not comply with the spec (at least I do think so ;-) there are servers and clients that ignore the "state" completely. See https://github.com/modelcontextprotocol/python-sdk/tree/main/examples/servers/simple-streamablehttp-stateless or https://github.com/modelcontextprotocol/java-sdk/blob/main/mcp-core/src/main/java/io/modelcontextprotocol/server/transport/HttpServletStatelessServerTransport.java.

lordofthejars:
I found this issue while creating an MCP Server to connect with IBM Bob. Bob doesn't do an init, it directly does a `tools/call`.  Moreover, doesn't send a reinitialization when receivers an HTTP 404. 

I noticed that after developing an MCP Server with Quarkus, that others MCP servers worked but not mine. After talking with Martin, we noticed this behaviour, but since other MCP servers works, means that basically  these other MCP servers are not following the MCP spec at all.

mkouba:
> I found this issue while creating an MCP Server to connect with IBM Bob. Bob doesn't do an init, it directly does a `tools/call`. Moreover, doesn't send a reinitialization when receivers an HTTP 404.

I think that it does init, but only once, and then all subsequent requests use the same session id, no matter what. Which is bad, because the server may terminate the session at any time, i.e. the server can be restarted and so all in-memory sessions are gone.

Issue #523, opened 2025-12-10
Guardrail for MCP tool


cescoffier:
Quarkus langchain4j just got tool guardrails. It would be great if we could apply the same set of annotation on tools exposed using the MCP server.

cescoffier:
TooInputGuardrails is not only about parameter validation, it can checks lots of other things.

They are kind of decorators, I initially thought about implementing them like these (but it would prevent the possibility to move them to langchain4j).

mkouba:
> TooInputGuardrails is not only about parameter validation, it can checks lots of other things.

Which things? All the examples in the docs are about params. And the `ToolInputGuardrail` interface even states _"Validates tool input parameters before execution."_.

For security, we support `@RolesAllowed("admin")`  and `@Authenticated` which is IMO more convenient.

For rate limiting, we could use [SmallRye Fault Tolerance](https://quarkus.io/guides/smallrye-fault-tolerance#adding-resiliency-rate-limits) 🤷.

> 
> They are kind of decorators, I initially thought about implementing them like these (but it would prevent the possibility to move them to langchain4j).

I see.

> @cescoffier so do you suggest to use the annotations from quarkus-langchain4j?

Pls could you answer this question ^?

sabob:
I view2 MCP tools as a different concept from langchain4j tools.
So I vote for having separate annotations in an mcp package.

cescoffier:
@mkouba For example, you can have input guardrails that remove PII from the parameters (or substitute/redact them to prevent PII Leakage) or detect toxic content in the generated arguments.

It is important to remember that these arguments are "inferred" by the LLM, there is no prior check on what was inferred. This is the critical spot where we can validate the semantics and safety of that inference, which goes beyond the structural checks handled by Bean Validation.

Regarding the annotations, I have no strong opinion, but I lean towards separate ones. I believe there will be a technical issue in reusing the same annotations because the "guardrail" implementations are likely to be fundamentally different. So, probably having your own dedicated annotations is better.

Also, in your case, interceptors are fine (there is no need to backport that in a CDI-less environment)

WDYT?

cescoffier:
@sberyozkin Speaking about that, did you look at: https://github.com/SAFE-MCP/safe-mcp ?

PR #556, opened 2026-01-09
Tool input/output guardrails

mkouba:
This is the first iteration of a POC for tool input/output guardrails (https://github.com/quarkiverse/quarkus-mcp-server/issues/523).

It was inspired by the API from [quarkus-langchain4j](https://github.com/quarkiverse/quarkus-langchain4j).

The `EmailFormatValidator` example from the javadoc of `io.quarkiverse.langchain4j.guardrails.ToolInputGuardrail` could be rewritten as:
```java
public class EmailFormatValidator implements ToolInputGuardrail {

        private static final Pattern EMAIL_PATTERN = Pattern.compile("^[A-Za-z0-9+_.-]+@[A-Za-z0-9.-]+$");

        @Override
        public void apply(ToolInputContext context) {
            String email = context.getArguments().getString("to");
            if (!EMAIL_PATTERN.matcher(email).matches()) {
                throw new ToolCallException("Invalid email format: " + email);
            }
        }
 }
```
There are a few differences:

- guardrails may support various execution models (worker thread, virtual thread, event loop); `@SupportedExecutionModels` can be used to supply the list of supported execution models; if a tool declares a guardrail with unsupported exec model the build fails
- a single annotation (`@ToolGuardrails`) is used to associate both input and output guardrails
- `ToolInputContext`/`ToolInputContext` provide access to specific APIs (such as `ToolInfo` or `McpConnection`)

TODO:

- [x] add support for tools added programmatically
- [ ] ~add support for "global" guardrails that are automatically associated with all tools; `@GlobalGuardrail` or something like that...~ global enablement probably does not make sense at this stage
- [x] docs

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/mcp/server/deployment/DotNames.java
+++ b/core/deployment/src/main/java/io/quarkiverse/mcp/server/deployment/DotNames.java
@@ -44,10 +44,12 @@ import io.quarkiverse.mcp.server.ResourceTemplateCompletionManager;
 import io.quarkiverse.mcp.server.ResourceTemplateManager;
 import io.quarkiverse.mcp.server.Roots;
 import io.quarkiverse.mcp.server.Sampling;
+import io.quarkiverse.mcp.server.SupportedExecutionModels;
 import io.quarkiverse.mcp.server.TextContent;
 import io.quarkiverse.mcp.server.TextResourceContents;
 import io.quarkiverse.mcp.server.Tool;
 import io.quarkiverse.mcp.server.ToolArg;
+import io.quarkiverse.mcp.server.ToolGuardrails;
 import io.quarkiverse.mcp.server.ToolManager;
 import io.quarkiverse.mcp.server.ToolResponse;
 import io.smallrye.common.annotation.Blocking;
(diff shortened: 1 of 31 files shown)


geoand:
Very interesting! I'll have a look on Monday

mkouba:
> > I’m not sure the client side is the right place for this since the client lacks context on the parameter semantics.
> 
> Could you elaborate what you mean by lacking parameter semantics? I struggle imagine a valid case when this would happen. If the tool has some "hidden" parameter semantics that are not obvious from the input schema and descriptions, then it's just a poorly written tool description. That would be like writing a Javadoc for a method and not mentioning some of the requirements for its parameters.

I'm not sure I completly understand this discussion but I believe that if an MCP server defines the input schema, description, validation requirements, etc. then it's up to the server to perform the validation. It simply cannot rely on the MCP client. It would be like if the server side of a web app relies on the client side validation.

And it's up to the server to choose the way the validation is perfomed. Someone prefers validation inside the tool method. Someone is used to HV constraints. This PR introduces just another way to perform the validation. And I believe that it's perfectly valid.

PR #602, opened 2026-02-10
resources: improve usability

### LINKED ISSUE #601: Resources: improve usability
- `ResourceContentsEncoder` should have access to the original `ResourceInfo`/`ResourceTemplateInfo` so that it can use the correct mimeType, etc.
- a `@Resource` method that returns `byte[]` should result in a response with single `BlobResourceContents`; mimeType and uri should be derived from `@Resource`
- a `@Resource` method that returns `String` should result in a response with single `TextResourceContents`; mimeType and uri should be derived from `@Resource`

mkouba:
- resolves #601


mkouba:
This is technically a breaking change (`String` and `byte` return types) but I think it makes perfect sense.

PR #640, opened 2026-02-27
Change version from 'dev' to 'true'

gastaldi:
https://docs.antora.org/antora/latest/playbook/content-source-version/#refname-as-version

### DIFF
--- a/docs/antora.yml
+++ b/docs/antora.yml
@@ -1,5 +1,5 @@
 name: quarkus-mcp-server
 title: MCP Server
-version: dev
+version: true
 nav:
   - modules/ROOT/nav.adoc


mkouba:
> https://docs.antora.org/antora/latest/playbook/content-source-version/#refname-as-version

I tried to read the link but still have no idea what it does :shrug:

gastaldi:
@mkouba the idea is that when you create a branch you don't need to manually change the nav.adoc to match the branch name (and avoid errors like https://github.com/quarkiverse/quarkiverse-docs/pull/264#issuecomment-3971407848), as the value of `true` does that for you

PR #667, opened 2026-03-06
core: support multiple McpServer bindings

### LINKED ISSUE #595: Make the McpServer annoration repeatable to that it's possible to bind a feature to multiple server configurations
First of all, we should investigate whether this would cause any compatibility issues.

mkouba:
- modify the API and adapt the Dev UI and tests
- add config flag to disable the support, i.e. enable compatibility mode
- resolves #595

### TODO
- [x] update docs

### DIFF
--- a/core/deployment/src/main/java/io/quarkiverse/mcp/server/deployment/FeatureMethodBuildItem.java
+++ b/core/deployment/src/main/java/io/quarkiverse/mcp/server/deployment/FeatureMethodBuildItem.java
@@ -1,8 +1,10 @@
 package io.quarkiverse.mcp.server.deployment;
 
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 
 import org.jboss.jandex.DotName;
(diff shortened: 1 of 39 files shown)


mkouba:
Also CC @Azquelt

phillip-kruger:
Using claude to review: 

qwc-mcp-tools.js — dialogRenderer dependency array updated, but other files didn't

In qwc-mcp-tools.js, _selectedServerName was correctly added to the dialogRenderer dependency array:
  ${dialogRenderer(() => this._renderToolInput(), [this._selectedTool, this._useRawJson, this._rawJsonInput, this._selectedServerName])}

  But in the other 5 files, the dialogRenderer calls do not include _selectedServerName in the dependency array. For example, in qwc-mcp-prompts.js line 189:
  ${dialogRenderer(() => this._renderPromptInput())}
  No dependency array at all. This means the dialog content might not re-render when the user changes the server selection in those components.
  
   Either add the dependency consistently across all files, or verify that Vaadin's dialog re-renders without it (it may depend on LitElement's reactive
  properties triggering a full re-render anyway, but it's inconsistent).

PR #674, opened 2026-03-11
core: fix pagination cursor

### LINKED ISSUE #672: Quarkus MCP server occasionally doesn't provide next pagination token
According to MCP specification[1] MCP server should support pagination for certain methods and include `nextCursor` field when there are more results. In case of Quarkus MCP server, this field sometimes is not included.

It seems, that this happens, then both of these are true:
- Tools* are declared in several files
- It is the second page of results (the first one always works properly)

How to reproduce:
1. `git clone -b mcp-no-next-token git@github.com:fedinskiy/reproducer.git reproducer && cd reproducer/server`
2. Start the server with small page size: ` mvn clean quarkus:dev -Phttp -Dquarkus.mcp.server.tools.page-size=2`
3. Initialize the client: `curl -i -H "Accept: application/json,text/event-stream" --data '{"jsonrpc": "2.0","id": 0,"method": "initialize","params":{"protocolVersion" : "2025-11-25","clientInfo" : {"name" : "custom","version" : "1.0"}}}' localhost:8080/mcp`
4. Request the initial page:
```
$ curl -s -H "Accept: application/json,text/event-stream" --data '{
  "jsonrpc" : "2.0",
  "id" : 1,
  "method" : "tools/list"
 }' localhost:8080/mcp | jq ".result.tools[].name, .result.nextCursor" 
"elicitation"
"sampled"
"MjAyNi0wMy0xMFQxNjoyMTo0OS41MTcwMDY5MzdaJCQkc2FtcGxlZA=="
```
5. Request the next page:
```
$  curl -s -H "Accept: application/json,text/event-stream" --data '{
   "jsonrpc" : "2.0",
   "id" : 2,
   "method" : "tools/list",
   "params" : {
     "cursor" : "MjAyNi0wMy0xMFQxNjoyMTo0OS41MTcwMDY5MzdaJCQkc2FtcGxlZA=="
   }

  }' localhost:8080/mcp | jq ".result.tools[].name, .result.nextCursor"
"unsampled"
"wait"
null # there is no next cursor, but we received only 4 tools
```
6. If we restart the server with bigger `page-size` (eg 13), then we can see, that there are 7 tools in total, so two two-value pages is definitely not enough:
```
$ curl -s -H "Accept: application/json,text/event-stream" --data '{                                                                                                                  "jsonrpc" : "2.0",
  "id" : 1,
  "method" : "tools/list"
 }' localhost:8080/mcp | jq ".result.tools[].name" 
"elicitation"
"sampled"
"unsampled"
"rootchecker"
"wait"
"filereader"
"meta"
```

* This issue may also affect other [three operations](https://modelcontextprotocol.io/specification/2025-11-25/server/utilities/pagination#operations-supporting-pagination) 

[1] https://modelcontextprotocol.io/specification/2025-11-25/server/utilities/pagination

mkouba:
- we cannot use the name to make the cursor resilient to concurrent modifications, instead use a snapshot timestamp
- fixes #672

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/mcp/server/FeatureManager.java
+++ b/core/runtime/src/main/java/io/quarkiverse/mcp/server/FeatureManager.java
@@ -39,7 +39,7 @@ public interface FeatureManager<INFO extends FeatureInfo> extends Iterable<INFO>
         boolean isMethod();
 
         /**
-         * @return the timestamp this feature was registered
+         * @return the timestamp when this feature was registered
          */
         Instant createdAt();
 
--- a/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/Cursor.java
+++ b/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/Cursor.java
@@ -1,34 +1,36 @@
 package io.quarkiverse.mcp.server.runtime;
 
-import java.nio.charset.StandardCharsets;
+import java.nio.ByteBuffer;
 import java.time.Instant;
-import java.util.Arrays;
 import java.util.Base64;
 
 /**
  * A composite opaque cursor.
  */
-record Cursor(Instant createdAt, String name) {
+record Cursor(Instant createdAt, Instant snapshotTimestamp) {
 
-    static final Cursor FIRST_PAGE = new Cursor(Instant.EPOCH, null);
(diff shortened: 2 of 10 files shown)


mkouba:
CC @fedinskiy

PR #703, opened 2026-03-26
OTel tracing implementation for the MCP server

brunobat:
This follows the OpenTelemetry semantic conventions for servers: https://opentelemetry.io/docs/specs/semconv/gen-ai/mcp/#server

Is only compatible with Quarkus 3.33+
If OTel is disabled either at build or runtime, the tracing overhead should be minimal.
I probably over tested this... We might remove some of them, depending on the review.


mkouba (commented):
I think that we should rework the CDI integration part (this pattern with a synthetic bean backed by a regular class bean is really weird and suboptimal) and potentially get rid of the NOOP `McpTracing` impl. We could still keep the NOOP `McpTracingSpan` though.

I also added a comment about moving the tracing logic into the `McpRequest.contextStart()` and `McpRequest.contextEnd()` callbacks.

I can't talk about the instrumenter part but it looks well covered by the tests (and I don't think we should remove them; does "over tested" even exist? ;-).

core/deployment/src/main/java/io/quarkiverse/mcp/server/deployment/McpServerProcessor.java
```diff
@@ -50,6 +50,7 @@
 import org.jboss.jandex.gizmo2.Jandex2Gizmo;
 import org.jboss.logging.Logger;
 
+import io.opentelemetry.api.OpenTelemetry;
```
mkouba: If the OTel dependency is optional, we cannot import the class directly. This also applies to `io.quarkus.opentelemetry.runtime.config.runtime.OTelRuntimeConfig`.

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/tracing/McpRequestInfo.java
```diff
@@ -0,0 +1,91 @@
...
+import io.quarkiverse.mcp.server.runtime.McpRequest;
+import io.quarkiverse.mcp.server.runtime.Messages;
+import io.vertx.core.json.JsonObject;
+
+public class McpRequestInfo {
+
+    private final McpMethod method;
+    private final JsonObject message;
+    private final McpRequest mcpRequest;
+    private final String networkTransport;
```
mkouba: This could be derived from the MCP transport, right? If so, then `mcpRequest.connection().initialRequest().transport()` could be used.

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/McpRequestImpl.java
```diff
@@ -17,6 +22,9 @@ public abstract class McpRequestImpl<CONNECTION extends McpConnectionBase> imple
     private final ManagedContext requestContext;
     private final CurrentIdentityAssociation currentIdentityAssociation;
 
+    // Tracing span - started by prepareTracing(), ended by contextEnd()
+    private McpTracingSpan tracingSpan;
```
mkouba: I think that this should be at least `volatile`.

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/McpRequestImpl.java
```diff
@@ -77,10 +102,29 @@ public void contextStart() {
                 securitySupport.setCurrentIdentity(currentIdentityAssociation);
             }
         }
+        // Store tracing request info in the Vert.x duplicated context
+        if (tracingSpan != null) {
+            McpRequestInfo ri = tracingSpan.requestInfo();
+            if (ri != null) {
+                io.vertx.core.Context ctx = Vertx.currentContext();
+                if (ctx != null) {
+                    ctx.putLocal(McpTracing.CONTEXT_KEY, ri.copy());
```
mkouba: Why do we actually store a copy? The object itself is immutable (except for the `responseInfo`) so a copy does not make a lot of sense to me.

Do we actually have to store something in the duplicated context? Couldn't we just store this information in the `McpRequest`?

core/deployment/src/main/java/io/quarkiverse/mcp/server/deployment/McpServerProcessor.java
brunobat: Fixed

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/tracing/McpRequestInfo.java
brunobat: I think I will store these values under the Transport enum.

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/McpRequestImpl.java
brunobat: Done.

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/McpRequestImpl.java
brunobat: yes, no need for copy here. I was being over zealous because of other issues related with context prop. Immutability saves the day here. 
Fixed.

brunobat:
@mkouba I've rewired the tracing instrumentation and the way I collect the tracing attributes.
I see that you did extensive changes in the IT tests. Do you want me to rewrite the tracing ones?

mkouba:
> I see that you did extensive changes in the IT tests. Do you want me to rewrite the tracing ones?

@brunobat Hm, which changes do you have in mind?

brunobat:
> > I see that you did extensive changes in the IT tests. Do you want me to rewrite the tracing ones?
> 
> @brunobat Hm, which changes do you have in mind?

I copied your IT tests back in the day and I enhanced them with OTel related wiring and assertions. In the meantime, you changed your IT tests. Should I rewrite mine accordingly?... Or leaving them like they are is fine.

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/ToolMessageHandler.java
```diff
@@ -77,7 +81,16 @@ Future<Void> toolsCall(JsonObject message, McpRequest mcpRequest) {
         try {
             Future<ToolResponse> fu = manager.execute(toolName,
                     new FeatureExecutionContext(message, mcpRequest));
-            return fu.compose(toolResponse -> mcpRequest.sender().sendResult(id, toolResponse),
+            return fu.compose(toolResponse -> {
+                if (toolResponse.isError()) {
+                    Context ctx = Vertx.currentContext();
+                    McpRequestInfo ri = ctx != null ? ctx.getLocal(McpTracing.CONTEXT_KEY) : null;
+                    if (ri != null) {
+                        ri.setResponseInfo(new McpResponseInfo(true, null, null));
```
mkouba: Couldn't we just use `McpRequest#setTracingErrorResponse()` here and get rid of the `McpTracing.CONTEXT_KEY` completely?

I mean, we could also change the `MessageHandler#handleFailure()` to accept the `McpRequest` and get rid of `MessageHandler#setTracingErrorResponse(int, String)` as well.

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/tracing/McpRequestInfo.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public McpRequestInfo(McpMethod method, JsonObject message, McpRequest mcpRequest, InitialRequest.Transport transport) {
+        this.method = method;
+        this.message = message;
+        this.mcpRequest = mcpRequest;
+        this.transport = transport;
+        this.responseInfo = new AtomicReference<>();
+    }
+
+    private McpRequestInfo(McpMethod method, JsonObject message, McpRequest mcpRequest, InitialRequest.Transport transport,
```
mkouba: This constructor is never used.

mkouba:
> > > I see that you did extensive changes in the IT tests. Do you want me to rewrite the tracing ones?
> > 
> > 
> > @brunobat Hm, which changes do you have in mind?
> 
> I copied your IT tests back in the day and I enhanced them with OTel related wiring and assertions. In the meantime, you changed your IT tests. Should I rewrite mine accordingly?... Or leaving them like they are is fine.

You mean the integration tests for STDIO, right? Yes, so we added the support for STDIO into `McpAssured`. Pls rewrite them, it's much easier now ;-)

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/tracing/McpRequestInfo.java
brunobat: yeah... Leftover.
Removed.

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/ToolMessageHandler.java
brunobat: Makes sense. Let me see what happens.

brunobat:
@mkouba I've simplified the context propagation, adapted the tests and also greatly increased the test assertions. Please make sure the asserted span names in the IT tests are the ones you would expect.

brunobat:
Btw, do you want me to squash the commits?

mkouba:
> Btw, do you want me to squash the commits?

Yes, we'll squash the commits once we're ready to merge ;-).

Give me pls a day or two to do the final review :pray:.

mkouba:
@brunobat FYI it seems that the MCP spec draft adds some docs about OpenTelemetry Trace Context Propagation conventions in https://modelcontextprotocol.io/specification/draft/basic#_meta

brunobat:
@mkouba We are propagating the context using the `"_meta"` attribute, but the span parent doesn't seem right in the tests. 
Current clients don't use OTel and I need to improve them.
We are also not linking to the transport server span, if present.

brunobat:
Not sure of how to handle HTTP propagation with batches. Needs research.
I've not implemented tracing on the stdio client. 
For web-sockets I think it would be possible but I haven't modified the client because the server side should work the same and...
The PR is getting really huge. 

Added docs and runtime config... I'm not sure if I'm using it in the right way.

test/src/main/java/io/quarkiverse/mcp/server/test/McpAssured.java
```diff
@@ -315,6 +315,16 @@ public interface Builder<BUILDER extends Builder<BUILDER>> {
...
 
+            /**
+             * Enables client-side MCP tracing. When set, the client creates CLIENT spans
+             * for each MCP request and injects trace context ({@code traceparent}, {@code tracestate})
+             * into the {@code _meta} field of outgoing messages.
+             *
+             * @param openTelemetry the OpenTelemetry instance to use for tracing
+             * @return self
+             */
+            BUILDER setTracing(io.opentelemetry.api.OpenTelemetry openTelemetry);
```
mkouba, line 326: We can't use `io.opentelemetry.api.OpenTelemetry` in the API - OTel is an optional dependency.

brunobat: For tests we can make that API not optional. no?

mkouba: Hm, I don't think so. I mean, the dependency is optional but since `McpAssured` is referencing the class directly it's not optional at all... And I don't think it's a good idea to force the users of `McpAssured` to always include the `opentelemetry-api` API.

brunobat: Are you thinking on promoting this to an independent project that anyone could use?... I think it would be cool.
Then, for this to work, I'need to add some kind of abstraction that will hide the `opentelemetry-api` dependency. Let me think about it.

mkouba: > Are you thinking on promoting this to an independent project that anyone could use?... I think it would be cool.

Do you mean `McpAssured`? Well, I think that it could be used standalone even now, although it brings the dependency on `quarkus-mcp-server-core`, because we share as many types as possible.

> Then, for this to work, I'need to add some kind of abstraction that will hide the `opentelemetry-api` dependency. Let me think about it.

In theory, we could just make the `io.opentelemetry:opentelemetry-api` non-optional and accept a few more additional dependencies :shrug:.

brunobat:
@mkouba I removed the uri from the resource spans to avoid high cardinality.
Squashed.

mkouba:
@brunobat Thanks!

PR #712, opened 2026-03-30
feat: calling support to sampling

### LINKED ISSUE #529: Add tool calling support to sampling
SEP-1577: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1577

Docs: https://modelcontextprotocol.io/specification/2025-11-25/client/sampling#sampling-with-tools

matheusandre1:
this Pr Close: #529

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/mcp/server/Content.java
+++ b/core/runtime/src/main/java/io/quarkiverse/mcp/server/Content.java
@@ -8,7 +8,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
  * A content provided to or from an LLM.
  */
 public sealed interface Content
-        permits TextContent, ImageContent, EmbeddedResource, AudioContent, ResourceLink {
+        permits TextContent, ImageContent, EmbeddedResource, AudioContent, ResourceLink, ToolUseContent, ToolResultContent {
 
     /**
      * @return the type of the content
@@ -75,6 +75,26 @@ public sealed interface Content
         throw new IllegalArgumentException("Not a resource link");
     }
 
+    /**
+     * Casts and returns this object as a tool use content, or throws an {@link IllegalArgumentException} if the content object
+     * does not represent a {@link ToolUseContent}.
+     *
+     * @return the tool use content
+     */
+    default ToolUseContent asToolUse() {
+        throw new IllegalArgumentException("Not a tool use");
+    }
+
+    /**
+     * Casts and returns this object as a tool result content, or throws an {@link IllegalArgumentException} if the content
+     * object does not represent a {@link ToolResultContent}.
+     *
+     * @return the tool result content
+     */
+    default ToolResultContent asToolResult() {
+        throw new IllegalArgumentException("Not a tool result");
+    }
+
     @JsonProperty("type")
     default String getType() {
         return type().toString().toLowerCase();
@@ -93,7 +113,9 @@ public sealed interface Content
         IMAGE,
         RESOURCE,
         AUDIO,
-        RESOURCE_LINK
+        RESOURCE_LINK,
+        TOOL_USE,
+        TOOL_RESULT
     }
 
 }
--- a/core/runtime/src/main/java/io/quarkiverse/mcp/server/SamplingMessage.java
+++ b/core/runtime/src/main/java/io/quarkiverse/mcp/server/SamplingMessage.java
@@ -1,60 +1,104 @@
 package io.quarkiverse.mcp.server;
 
+import java.util.List;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 /**
- * @param role the sender or recipient of messages and data in a conversation (must not be {@code null})
- * @param content the message content (must not be {@code null})
- *
  * @see TextContent
  * @see ImageContent
+ * @see ToolUseContent
+ * @see ToolResultContent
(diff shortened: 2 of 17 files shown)


transports/http/deployment/src/test/java/io/quarkiverse/mcp/server/test/sampling/SamplingTest.java
```diff
@@ -86,6 +141,25 @@ Uni<String> samplingFoo(Sampling sampling) {
...
+        @Tool(description = "A tool that is using sampling with tools...")
+        Uni<String> samplingWithTools(Sampling sampling) {
+            SamplingRequest samplingRequest = sampling.requestBuilder()
+                    .setMaxTokens(100)
+                    .addMessage(SamplingMessage.withUserRole("What's the weather?"))
+                    .addTool(toolManager.getTool("weather_lookup"))
+                    .setToolChoice(new ToolChoice(ToolChoice.Mode.REQUIRED))
+                    .build();
+            return samplingRequest.send().map(sr -> {
+                ToolUseContent toolUse = sr.content().asToolUse();
```
mkouba, line 158: So if I understand it correctly this PR does not try to solve the "tool execution" part, i.e. it's up to the MCP tool author to process the `tool_use` request from the server. I.e. the MCP tool author would need to create the `ToolResultContent` manually, in a subsequent/nested sampling request.

core/runtime/src/main/java/io/quarkiverse/mcp/server/Content.java
```diff
@@ -8,7 +8,7 @@
  * A content provided to or from an LLM.
  */
 public sealed interface Content
-        permits TextContent, ImageContent, EmbeddedResource, AudioContent, ResourceLink {
+        permits TextContent, ImageContent, EmbeddedResource, AudioContent, ResourceLink, ToolUseContent, ToolResultContent {
```
mkouba, line 11: This does not reflect the `ContentBlock` schema https://modelcontextprotocol.io/specification/2025-11-25/schema#contentblock. And I think that we should try to keep the types in sync.

core/runtime/src/main/java/io/quarkiverse/mcp/server/SamplingMessage.java
```diff
@@ -1,60 +1,104 @@
...
- * @param content the message content (must not be {@code null})
- *
  * @see TextContent
  * @see ImageContent
+ * @see ToolUseContent
+ * @see ToolResultContent
  * @see SamplingRequest
  */
-public record SamplingMessage(Role role, Content content) {
+public final class SamplingMessage {
```
mkouba, line 14: I think that we will need to introduce an analogy for [`SamplingMessageContentBlock`](https://modelcontextprotocol.io/specification/2025-11-25/schema#samplingmessagecontentblock).

Also we should try to keep the backward compatibility. So I think that we should change the canonical constructor to accept a list of `SamplingContent` (see my comment above).  This change is not fully source-compatible if anyone uses record patterns, regular construction via the convenience constructor is fine.

core/runtime/src/main/java/io/quarkiverse/mcp/server/ToolChoice.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+public record ToolChoice(Mode mode) {
+
+    public ToolChoice {
+        if (mode == null) {
+            throw new IllegalArgumentException("mode must not be null");
+        }
+    }
+
+    public enum Mode {
```
mkouba, line 18: Maybe add some javadoc?

transports/http/deployment/src/test/java/io/quarkiverse/mcp/server/test/sampling/SamplingTest.java
mkouba: I think that we could add something to the `io.quarkiverse.mcp.server.ToolManager` to help users "execute" a tool. Except that there is a risk of nested calls and endless loops (a tool can call itself) :thinking:.

mkouba: Hm, the more I think about this, the more complications I can see. But the need for manual parsing of tool arguments is IMO very inconvenient.

mkouba:
@matheusandre1 BTW it seems that [Sampling is deprecated](https://modelcontextprotocol.io/specification/draft/client/sampling) in the latest DRAFT-2026-v1, so I'm not really sure it's worth the effort :shrug:.

matheusandre1:
@mkouba , Thank you for your patience and review, so would it be better to close this pull request?

mkouba:
> @mkouba , Thank you for your patience and review, so would it be better to close this pull request?

I'm not sure to be honest. I think that we could merge this PR if we address my previous comments, and leave the "tool execution" part up to the users. But that looks like an incomplete API.

@matheusandre1 Do you need this feature yourself? If not, I'd wait for more feedback from users.

matheusandre1:
Sorry for the delay; the issue I wanted this for was moved out of the planning stage, so it's no longer necessary. Thanks for your patience.

mkouba:
> Sorry for the delay; the issue I wanted this for was moved out of the planning stage, so it's no longer necessary. Thanks for your patience.

No problem at all. Thanks.

PR #737, opened 2026-04-13
Fix callout rendering in custom encoders guide

rolfedh:
## Summary

- Remove blank line between numbered list and code block in `guides-custom-encoders.adoc` so the code block stays attached to the list and the callout renders correctly

## Test plan

- [ ] Verify the callout `<1>` renders correctly in the "Encoder Selection for Tools" section

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rolfedh:
Incorrect.

PR #738, opened 2026-04-13
Fix logging format placeholders in doc examples

### LINKED ISSUE #739: Doc: logging examples use wrong format placeholders
## Description

The logging examples in the documentation use SLF4J-style `{}` placeholders, but `McpLog` delegates to JBoss Logger `f`-variant methods (`infof`, `debugf`, `errorf`) which use `java.util.Formatter` syntax. The `send()` method also uses `String.formatted()`. The correct placeholders are `%s`, `%d`, etc.

As written, the examples will print literal `{}` instead of interpolating the arguments.

## Affected files

- `docs/modules/ROOT/pages/guides-logging-progress-cancellation.adoc` (6 instances)
- `docs/modules/ROOT/pages/guides-implementing-prompts.adoc` (1 instance)

## Evidence from source

- [`McpLog.info()`](https://github.com/quarkiverse/quarkus-mcp-server/blob/main/core/runtime/src/main/java/io/quarkiverse/mcp/server/McpLog.java#L51) Javadoc references `Logger#infof(String, Object...)`
- [`McpLogImpl.info()`](https://github.com/quarkiverse/quarkus-mcp-server/blob/main/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/McpLogImpl.java#L53-L56) calls `logger.infof(format, params)` and `send()` which uses `format.formatted(params)`

## Backport request

Please backport the fix to the `1.10.x` and `1.11.x` branches, as downstream distributions (including IBM builds) consume these versions.

## Reported by

IBM QE, via downstream documentation testing.

rolfedh:
McpLog delegates to JBoss Logger f-variant methods (infof, debugf, errorf)
which use java.util.Formatter syntax. The doc examples incorrectly used
SLF4J-style `{}` placeholders instead of `%s`/`%d`.

Changes across 2 files, 7 instances:
- `{}` → `%s` for String arguments
- `{}` → `%d` for integer arguments (`result.length()`, loop counter `i`)
- `{}%` → `%d%%` for the percentage format (escaped literal `%`)

* Fixes #739

Please backport to `1.10.x` and `1.11.x` for downstream IBM builds.

### DIFF
--- a/docs/modules/ROOT/pages/guides-implementing-prompts.adoc
+++ b/docs/modules/ROOT/pages/guides-implementing-prompts.adoc
@@ -141,7 +141,7 @@ import io.quarkiverse.mcp.server.McpLog;
 
 @Prompt(description = "Prompt with logging")
 PromptMessage loggedPrompt(String input, McpLog log) {
-    log.info("Processing prompt with input: {}", input);
+    log.info("Processing prompt with input: %s", input);
     return PromptMessage.withUserRole(new TextContent("Processing: " + input));
 }
 ----
--- a/docs/modules/ROOT/pages/guides-logging-progress-cancellation.adoc
+++ b/docs/modules/ROOT/pages/guides-logging-progress-cancellation.adoc
@@ -24,12 +24,12 @@ public class MyTools {
 
     @Tool(description = "Process data with logging")
     String processData(String input, McpLog log) {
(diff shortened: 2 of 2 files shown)


mkouba:
Great catch! I think that we should also clarify the API, i.e. mention the placeholders in the `McpLog`.

Issue #308, opened 2025-06-23
IllegalStateException sometimes appears when tools perform logging


jmartisk:
When calling tools that perform logging, over the new Streamable HTTP transport, I sometimes see the execution failing, this appears on the server

```
java.lang.IllegalStateException: Response has already been written
	at io.vertx.core.http.impl.Http1xServerResponse.write(Http1xServerResponse.java:773)
	at io.vertx.core.http.impl.Http1xServerResponse.write(Http1xServerResponse.java:351)
	at io.quarkus.vertx.http.runtime.filters.AbstractResponseWrapper.write(AbstractResponseWrapper.java:186)
	at io.quarkiverse.mcp.server.sse.runtime.StreamableHttpMcpMessageHandler$HttpMcpRequest.send(StreamableHttpMcpMessageHandler.java:491)
	at io.quarkiverse.mcp.server.runtime.McpLogImpl.send(McpLogImpl.java:41)
	at io.quarkiverse.mcp.server.runtime.McpLogImpl.debug(McpLogImpl.java:60)
	at logging_mcp_server.debug(logging_mcp_server.java:25)
	at logging_mcp_server_debug_Invoker_Qdq7gL9bx1yjHIPYmfk5ko4KvD0.invoke(Unknown Source)
	at logging_mcp_server_debug_LazyInvoker_Qdq7gL9bx1yjHIPYmfk5ko4KvD0.invoke(Unknown Source)
	at io.quarkiverse.mcp.server.runtime.FeatureManagerBase$FeatureMetadataInvoker.call(FeatureManagerBase.java:300)
	at io.quarkiverse.mcp.server.runtime.FeatureManagerBase$1.call(FeatureManagerBase.java:69)
	at io.quarkiverse.mcp.server.runtime.FeatureManagerBase$1.call(FeatureManagerBase.java:66)
	at io.quarkiverse.mcp.server.runtime.FeatureManagerBase$3.call(FeatureManagerBase.java:232)
	at io.quarkiverse.mcp.server.runtime.FeatureManagerBase$3.call(FeatureManagerBase.java:228)
	at io.vertx.core.impl.ContextImpl.lambda$executeBlocking$4(ContextImpl.java:192)
	at io.vertx.core.impl.ContextInternal.dispatch(ContextInternal.java:270)
	at io.vertx.core.impl.ContextImpl$1.execute(ContextImpl.java:221)
[...]

PR #744, opened 2026-04-16
core: fix log level in McpLog#error(Throwable) overload

rolfedh:
## Problem

`McpLogImpl.error(Throwable t, String format, Object... params)` calls `logger.infof(t, format, params)` instead of `logger.errorf(...)`. As a result, when an MCP feature logs a throwable via this overload:

- The **local JBoss Logger output** is emitted at **INFO** (and would be suppressed in environments that raise the log threshold above INFO).
- The **MCP notification sent to the client** is correctly at **ERROR**.

The two diverge, and the local error context on the server is silently downgraded — including under the production-default root-level thresholds that filter INFO.

## Fix

One-character change: `logger.infof` → `logger.errorf` in [`McpLogImpl.java:72`](https://github.com/quarkiverse/quarkus-mcp-server/blob/main/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/McpLogImpl.java#L72).

## Evidence this is a typo, not intent

The other three format-taking methods all align the local logger call with their level:

| Method | Local call | MCP notification |
|---|---|---|
| `info(format, params)` | `logger.infof` | `LogLevel.INFO` |
| `debug(format, params)` | `logger.debugf` | `LogLevel.DEBUG` |
| `error(format, params)` | `logger.errorf` | `LogLevel.ERROR` |
| `error(t, format, params)` **(before)** | `logger.infof` ❌ | `LogLevel.ERROR` |
| `error(t, format, params)` **(after)** | `logger.errorf` ✅ | `LogLevel.ERROR` |

`git blame` shows the line has been `log.infof(t, ...)` / `logger.infof(t, ...)` since before the streamable-transport refactor in #308 — a copy-paste from the `info(String, Object...)` method directly above it.

## Tests

I did not add a test. The existing `LoggingTest` suite asserts MCP notifications sent to the client — not local JBoss Logger output — so a notification-level assertion wouldn't catch this class of bug. Verifying the local level requires wiring a `java.util.logging` / JBoss Logger handler capture, which is more infrastructure than the fix itself. Happy to add one in a follow-up if maintainers prefer.

## Backport

Please backport to `1.10.x` and `1.11.x` for downstream IBM builds.

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/McpLogImpl.java
+++ b/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/McpLogImpl.java
@@ -69,7 +69,7 @@ class McpLogImpl implements McpLog {
 
     @Override
     public void error(Throwable t, String format, Object... params) {
-        logger.infof(t, format, params);
+        logger.errorf(t, format, params);
         send(LogLevel.ERROR, format, params);
     }
 


mkouba:
@rolfedh Thanks for the fix.

> Please backport to 1.10.x and 1.11.x for downstream IBM builds.

1.11.x should not be used in any downstream IBM build :thinking:.

Issue #570, opened 2026-01-23
Add integration-tests where a secured MCP server can be launched and tested


sberyozkin:
I'm not sure it is a valid request, I'd like to have an integration test where we can have an MCP server launched, and I can send some tokens to it to verify how they are protected, to check OIDC protected resource metadata.

I see https://github.com/quarkiverse/quarkus-mcp-server/tree/main/transports/http/integration-tests, is it where it should be done ?

mkouba:
@sberyozkin Hm, couldn't we just add an integration test to the [`secure-mcp-sse-server`](https://github.com/quarkiverse/quarkus-mcp-server/tree/main/samples/secure-mcp-sse-server) sample app?

mkouba:
@sberyozkin Should we close this issue?

sberyozkin:
@mkouba Yes, thanks, let's do it

sberyozkin:
In addition to #771, the following PRs also were merged: #772, #775, #780

PR #761, opened 2026-04-22
Add a first set of MCP Authorization integration tests

sberyozkin:
Related to #570.

This PR adds a first set of MCP authorization tests that verify key MCP authorization requirements for MCP servers and are structured as follows:

* Protected MCP server is deployed
* JAX-RS test endpoint is provided only to help MCP test clients to get hold of the authorization code flow tokens
* One test verifies that by using a token with a correct audience produces expected response
* One test verifies that by using a token without a correct audience produces 401
* Test without a token verifies 2 things 1) that 401 is returned and 2) WWW-Authenticate contains a `resource_metadata` property that can be used to fetch MCP server metadata, get an authorization server url from it and that this url is equal to the Keycloak url protecting this MCP server

To get these tests working, McpAssured had to be updated to accept connection failure assertion consumers, following the same pattern as the one for the successful connections

I'm not marking this PR to fix the linked issue yet, I'd like to review what other test scenarios can be tested.

I was working with Claude on this PR.

CC @mkouba @cescoffier


sberyozkin:
@mkouba Hi Martin, I updated the PR to have these tests run only if a `-Dtest-containers` property is enabled as otherwise it fails on Windows.
But I may ask you for some help to have https://github.com/quarkiverse/quarkus-mcp-server/blob/main/.github/workflows/build.yml#L37 updated as you prefer to have these tests run

test/src/main/java/io/quarkiverse/mcp/server/test/McpAssured.java
```diff
@@ -467,6 +467,16 @@ public interface Builder extends McpTestClient.Builder<Builder> {
...
 
+            /**
+             * Expects a connect failure and uses the provided assert function to verify the failure response.
+             * <p>
+             * Note that the client will not be connected after the invocation of {@link McpStreamableTestClient#connect()}.
+             *
+             * @param assertFunction
+             * @return self
+             */
+            Builder setExpectConnectFailure(Consumer<ConnectFailureResponse> assertFunction);
```
mkouba, line 481: Maybe we could change the `setExpectConnectFailure()` to:
```java
default Builder setExpectConnectFailure()  {
   return setExpectConnectFailure(r -> {});
}
```
and replace the `expectConnectFailure` boolean flag in `McpStreamableTestClientImpl` with the consumer?

mkouba (commented):
Looks good. Just added one minor comment.

mkouba:
> @mkouba Hi Martin, I updated the PR to have these tests run only if a `-Dtest-containers` property is enabled as otherwise it fails on Windows. But I may ask you for some help to have https://github.com/quarkiverse/quarkus-mcp-server/blob/main/.github/workflows/build.yml#L37 updated as you prefer to have these tests run

So it's expected to fail on Windows?

test/src/main/java/io/quarkiverse/mcp/server/test/McpAssured.java
sberyozkin: @mkouba This is done, thanks for the idea, looks neater

sberyozkin:
@mkouba It does work on Windows in Quarkus, but I'm not sure how the docker environment is setup there.
@gsmet, @cescoffier Can you please clarify how to get the tests that require loading a container pass on Windows in CI ?

mkouba (approved):

mkouba:
> @mkouba It does work on Windows in Quarkus, but I'm not sure how the docker environment is setup there. 

Ok, I see. We could configure the Windows CI in a follow-up pull request.

sberyozkin:
@mkouba Sure, the problem is that the tests don't run right now, unless `-Dtest-containers` is passed in CI

sberyozkin:
I mean, I can enable them for now, but all CIs for all PRs will start failing on Windows as a result

sberyozkin:
May be we can just disable them on Windows ? Not sure how to do it though

mkouba:
> May be we can just disable them on Windows ? Not sure how to do it though

I will take a look, don't remember from the top of my head but I used a similar config in the past...

sberyozkin:
Great, thanks @mkouba

PR #765, opened 2026-04-24
docs(multiple-servers): add explicit kebab-case IDs to all section headings

rolfedh:
## Summary

Replace renderer-dependent auto-generated anchors with explicit ``[id="..."]`` attributes on every section heading in `docs/modules/ROOT/pages/guides-multiple-servers.adoc`, so the anchors are stable across Asciidoctor, DITA, and any other downstream renderer.

No content changes — IDs only.

PR #766, opened 2026-04-28
docs: add tooling to enforce explicit heading IDs (1/2)

rolfedh:
## Context

This PR implements the tooling agreed in the "Decisions to confirm" section of [quarkus-langchain4j#2401](https://github.com/quarkiverse/quarkus-langchain4j/issues/2401), as confirmed by @mkouba in [#765 (comment)](https://github.com/quarkiverse/quarkus-mcp-server/pull/765#issuecomment-4337982408).

Goal: every section heading in `docs/modules/ROOT/pages/` should have an explicit kebab-case `[id="..."]` so anchors stay stable across renderer changes.

Per Decision 6 (two PRs), this is **PR 1 of 2** — tooling only. The follow-up PR will run `--fix` across the doc tree and commit the resulting ID lines.

## What's here

- `docs/scripts/check-heading-ids.py` — Python 3 stdlib only, three modes:
  - `--check` (CI default): exits 1 on violations
  - `--fix`: applies IDs in place
  - `--self-test`: inline assertions for the derivation regex
- `.github/workflows/docs-id-check.yml` — runs `--check` on PRs that touch `docs/modules/ROOT/pages/**`, the script, or the workflow file.

## How it maps to the six decisions

| Decision in #2401 | Implementation |
|---|---|
| 1. Kebab-case explicit IDs framework-wide | Derivation rule + `--check` form-validation (`^[a-z0-9]+(-[a-z0-9]+)*$`) |
| 2. `[id="..."]` form, not `[[anchor]]` | Script emits canonical form and migrates legacy `[[anchor]]` → `[id="anchor"]` |
| 3. Every heading, not just cross-referenced | Script processes every level-1+ heading (`==` to `======`); level-0 page title is skipped because Antora derives the page anchor from the filename |
| 4. Script in `docs/scripts/` | `docs/scripts/check-heading-ids.py` |
| 5. Standalone `docs-id-check.yml` workflow | `.github/workflows/docs-id-check.yml` |
| 6. Two PRs (tooling, then mechanical apply) | This is PR 1 |

## Script behaviors worth noting

- **Block-fence masking**: skips heading-shaped lines inside `----`, `....`, `++++`, `////`, and `====` blocks, so source-listing and admonition contents aren't accidentally tagged.
- **Attribute-stack walking**: handles stacked block attributes (e.g., `[role="..."]` above the heading) and recognizes both `[id="..."]` and shorthand `[#id]`.
- **Non-destructive**: `--fix` never modifies an existing `[id="..."]` or `[#id]`, even if it's not kebab-case — rewriting a public anchor would silently break inbound links. Non-kebab existing IDs surface via `--check` for human resolution.
- **Idempotent**: re-running `--fix` after a clean `--check` produces no diff.
- **Scope**: `docs/modules/ROOT/pages/`, excluding `pages/includes/` (auto-generated config docs).

## Verification

- `--self-test`: 24 inline assertions pass (derivation, kebab regex, fence detection, ID extraction).
- `--check` against `main`: exits 1 with 468 violations — 464 missing IDs + 4 legacy `[[anchor]]` lines in `reference-configuration.adoc`, `reference-filters-and-checks.adoc`, `reference-parameter-types.adoc`, `reference-return-types.adoc`. These will be fixed by PR 2 of 2.

## Follow-up

PR 2 will extend the existing `docs-explicit-heading-ids` branch (currently 1 commit ahead of `main` with the proof-of-concept on `guides-multiple-servers.adoc`) with the bulk apply.


rolfedh:
Heads up: the new `docs-id-check` workflow is failing on this PR — that is expected.

The workflow runs `--check` against the whole `docs/modules/ROOT/pages/` tree, which today has ~468 headings without explicit `[id="..."]`. PR 1 only introduces the tooling; PR 2 (the mechanical apply, coming next per Decision 6 in [quarkus-langchain4j#2401](https://github.com/quarkiverse/quarkus-langchain4j/issues/2401)) will run `--fix` and bring the count to zero.

Once both PRs land, the check stays green for all future PRs.

The first failing run, for reference: https://github.com/quarkiverse/quarkus-mcp-server/actions/runs/25078638373/job/73477939412?pr=766

.github/workflows/docs-id-check.yml
```diff
@@ -0,0 +1,23 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
```
mkouba, line 19: @gastaldi Is it ok to use this GitHub action in quarkiverse projects?

docs/scripts/check-heading-ids.py
```diff
@@ -0,0 +1,357 @@
...
+preceded by an explicit `[id="kebab-case-id"]` block attribute.
+
+Usage:
+    python docs/scripts/check-heading-ids.py --check       # CI-friendly, exit 1 on violations
+    python docs/scripts/check-heading-ids.py --fix         # rewrite files in place
+    python docs/scripts/check-heading-ids.py --self-test   # run inline assertions
+
+Scope: docs/modules/ROOT/pages/ excluding pages/includes/ (auto-generated).
+
+Generated by Claude Code.
```
mkouba, line 16: I'm definitely not against using AI tools in this project, but keep in mind that [Quarkus AI Usage Policy](https://github.com/quarkusio/quarkus/blob/main/AI_POLICY.md) applies here as well. In other words, I hope you understand this Python script, and take the responsibility for its validity and correctness.

.github/workflows/docs-id-check.yml
gastaldi: Yes, that's fine

rolfedh:
Thanks @mkouba — I want to respond directly to the AI Usage Policy point, because I think it's the right call here.

I studied this script carefully, and the honest truth is I can't yet vouch for its correctness to the standard the policy asks for. It's a bulk **rewriter** over AsciiDoc — the `--fix` path edits every heading across the doc tree — and that's exactly the kind of tool where partial understanding isn't good enough. The edge-case surface (the ID-derivation pipeline, the delimited-block handling, the in-place mutation) is real, and I'd be asserting an ownership I don't actually have if I pushed it through. So rather than do that, I'd rather withdraw it.

Closing this one. The script is still on the branch if anyone wants to pick it up and genuinely own it — happy for it to be a starting point — but I don't want it merged on my name without the confidence the policy (rightly) expects.

Appreciate you holding the line on this; it's the correct bar.

PR #775, opened 2026-05-06
Add a handler to intercept security failures after a Streamable HTTP is established

sberyozkin:
Myself and Claude tried many things :-), this one looks not bad to me - the security exception may be thrown at the OIDC level for various reasons after a successful connection, so the proposed handler would correctly convert such failures to JSON-RPC responses, thus avoiding MCP clients hanging.

The proposed test was nearly working even before the PR, I just did not realize the MCP client was hanging - no security issue at all - all works as expected at the security level with respect to this test. CC @cescoffier

### DIFF
--- a/transports/http/deployment/src/main/java/io/quarkiverse/mcp/server/http/deployment/HttpMcpServerProcessor.java
+++ b/transports/http/deployment/src/main/java/io/quarkiverse/mcp/server/http/deployment/HttpMcpServerProcessor.java
@@ -75,6 +75,10 @@ public class HttpMcpServerProcessor {
                     .withRouteCustomizer(recorder.addBodyHandler(bodyHandler.getHandler()))
                     .withRequestHandler(recorder.createMcpEndpointHandler(endpoint.serverName))
                     .build());
+            routes.produce(RouteBuildItem.newFrameworkRoute(endpoint.mcpPath)
+                    .withRequestHandler(recorder.createAuthFailureHandler())
+                    .asFailureRoute()
+                    .build());
             // SSE/HTTP transport
             routes.produce(RouteBuildItem.newFrameworkRoute(endpoint.ssePath)
                     .withRequestHandler(recorder.createSseEndpointHandler(endpoint.mcpPath, endpoint.serverName))
--- a/transports/http/runtime/src/main/java/io/quarkiverse/mcp/server/http/runtime/HttpMcpServerRecorder.java
+++ b/transports/http/runtime/src/main/java/io/quarkiverse/mcp/server/http/runtime/HttpMcpServerRecorder.java
@@ -15,6 +15,7 @@ import java.util.function.Function;
 import org.eclipse.microprofile.config.ConfigProvider;
 import org.jboss.logging.Logger;
 
+import io.quarkiverse.mcp.server.JsonRpcErrorCodes;
 import io.quarkiverse.mcp.server.http.runtime.config.McpHttpServersBuildTimeConfig;
 import io.quarkiverse.mcp.server.http.runtime.config.McpHttpServersRuntimeConfig;
 import io.quarkiverse.mcp.server.runtime.ConnectionManager;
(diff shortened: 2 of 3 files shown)


transports/http/deployment/src/main/java/io/quarkiverse/mcp/server/http/deployment/HttpMcpServerProcessor.java
```diff
@@ -67,14 +70,22 @@ void serverNames(McpHttpServersBuildTimeConfig config, BuildProducer<ServerNameB
...
             BodyHandlerBuildItem bodyHandler,
+            Capabilities capabilities,
             BuildProducer<RouteBuildItem> routes) {
         for (McpServerEndpoint endpoint : mcpServerEndpoints.getEndpoints()) {
             // Streamable HTTP transport
             routes.produce(RouteBuildItem.newFrameworkRoute(endpoint.mcpPath)
                     .withRouteCustomizer(recorder.addBodyHandler(bodyHandler.getHandler()))
                     .withRequestHandler(recorder.createMcpEndpointHandler(endpoint.serverName))
                     .build());
+            if (capabilities.isPresent(Capability.SECURITY)) {
```
mkouba: `quarkus-security` should be always present - it's a dependency of the core module.

transports/http/runtime/src/main/java/io/quarkiverse/mcp/server/http/runtime/HttpMcpServerSecurityRecorder.java
```diff
@@ -0,0 +1,57 @@
...
+import io.quarkus.security.UnauthorizedException;
+import io.vertx.core.Handler;
+import io.vertx.core.http.HttpHeaders;
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.web.RoutingContext;
+
+@Recorder
+public class HttpMcpServerSecurityRecorder {
+
+    public Handler<RoutingContext> createAuthFailureHandler() {
```
mkouba: I'd move this into the `HttpMcpServerRecorder`.

sberyozkin:
Hi @mkouba Have another look please

transports/http/runtime/src/main/java/io/quarkiverse/mcp/server/http/runtime/HttpMcpServerRecorder.java
```diff
@@ -103,6 +108,47 @@ public void handle(RoutingContext ctx) {
...
+                    }
+                    if (!ctx.request().isEnded()) {
+                        ctx.request().resume();
+                    }
+                    ctx.request().body().onComplete(ar -> {
+                        Object id = null;
+                        if (ar.succeeded() && ar.result() != null) {
+                            try {
+                                id = new JsonObject(ar.result()).getValue("id");
+                            } catch (Exception ignored) {
```
mkouba: What kind of exception do you expect here?

sberyozkin: @mkouba Sorry, when I was looking at this code, I assumed Claude saw a similar pattern elsewhere. But now it is apologetic after I asked why it was added :-). This is silly, I'd have written it faster myself, oh well

sberyozkin: This catch block has gone

mkouba (approved):
Looks good. @sberyozkin do you want to wait for feedback from @cescoffier?

sberyozkin:
Thanks @mkouba, I thought we'd have a call with Clement where I'd go through all of `authorization-integration-tests` and discuss various MCP Authorization spec requirements  that are applicable to the MCP server part. This particular PR just makes sure the security errors returned after the connection do not cause MCP clients to hang and get the security error details

mkouba:
> Thanks @mkouba, I thought we'd have a call with Clement where I'd go through all of `authorization-integration-tests` and discuss various MCP Authorization spec requirements that are applicable to the MCP server part. This particular PR just makes sure the security errors returned after the connection do not cause MCP clients to hang and get the security error details

Ok, so I'll merge this pull request and potential issues can be addressed in a follow-up.

PR #776, opened 2026-05-07
core: fix tool argument conversion

### LINKED ISSUE #774: @ToolArg ignores the configured ObjectMapper for scalar parameter types / default input-schema generator doesn't reflect the serialized JSON shape
**Affected version:** `io.quarkiverse.mcp:quarkus-mcp-server-core` `1.12.0` (behavior is the same on `main` at time of writing).

## Summary

>In short, it's not possible to use a type of e.g. `LocalDate` for a `ToolArg`, nor any other types where you would want it to serialize as a string.

When a tool method declares a `@ToolArg` parameter whose Java type is a custom scalar (i.e. not `String`/`Number`/`Boolean`/`Map`/`List`/enum) that is serialized by Jackson as a single JSON string via a registered `JsonSerializer`/`JsonDeserializer`, two things go wrong:

1. **Runtime:** the argument is rejected before the registered `ObjectMapper` gets a chance to deserialize it. `FeatureArgument.expectedJsonTypeFrom(Type)` maps unknown types to `JsonType.OBJECT`, so a JSON string value fails `FeatureArgument.isValid(val)` with `value does not match the expected JSON type: object`. Even if validation passed, `FeatureManagerBase.handleParam` only invokes `mapper.convertValue(...)` for `Map` and `List` values — a raw `String` (or `Number`) targeting a custom type is passed through unchanged, causing a `ClassCastException` when the generated invoker assigns it to the parameter.
2. **Schema:** `DefaultSchemaGenerator` delegates to `victools/jsonschema-generator`, which reflects on the Java class structure rather than consulting Jackson. For a type like `MyDto` (which serializes as a single string via a custom `JsonSerializer`), the emitted input schema describes the Java fields of `MyDto`, not `"type": "string"`. Clients therefore see a misleading/incorrect schema.

The only current workaround is to declare every such parameter as `String` and re-parse it manually inside the tool method, which defeats the point of having a project-wide Jackson module.

## Repro

Project registers a Jackson `SimpleModule` (auto-discovered via `META-INF/services/com.fasterxml.jackson.databind.Module`) for a value-object type `MyDto` that is serialized to/from a single JSON string:

```java
public class MyModule extends SimpleModule {
  @Override
  public void setupModule(SetupContext context) {
    var s = new SimpleSerializers();
    s.addSerializer(new MyDtoSerializer());       // MyDto -> JSON string
    context.addSerializers(s);
    var d = new SimpleDeserializers();
    d.addDeserializer(MyDto.class, new MyDtoDeserializer()); // JSON string -> MyDto
    context.addDeserializers(d);
  }
}
```

Tool method:

```java
@Tool(description = "...")
public Uni<Result> lookup(
    @ToolArg(description = "The dto to look up.") MyDto dto) {
  // ...
}
```

Call with JSON:

```json
{ "name": "lookup", "arguments": { "dto": "some-string-value" } }
```

**Expected:** `dto` is deserialized into a `MyDto` via the registered `MyDtoDeserializer`; the generated input schema describes `dto` as `"type": "string"` (optionally with `description` from `@ToolArg`).

**Actual:**
- Input schema: `"dto": { "type": "object", "properties": { …internal fields of MyDto… } }`
- Call returns `INVALID_PARAMS` — `Invalid argument [dto] - value does not match the expected JSON type: object`
- If you change the tool param type to `String` and parse by hand it works, but `@JsonSerialize`/`@JsonDeserialize` and `Module`-provided serializers become dead code.

## Relevant code paths (1.12.0)

- `core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/FeatureArgument.java` — `expectedJsonTypeFrom(Type)` defaults every unknown type to `OBJECT`; `isValid(Object)` uses that mapping.
- `core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/FeatureManagerBase.java` — `handleParam(...)` only calls `mapper.convertValue(...)` for `Map`/`List` inputs. String/Number inputs are never routed through the `ObjectMapper`, so registered `JsonDeserializer`s are bypassed.
- `core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/DefaultSchemaGenerator.java` — `generateSchema(Type, String, String)` delegates to `victools` `SchemaGenerator`, which doesn't consult the `ObjectMapper`, so it emits the Java-side shape rather than the Jackson-serialized shape.

## Proposed fix

The underlying pattern is: **compare the JSON value's runtime type to the target Java type. If they're the same, pass through; if they differ, always route through the `ObjectMapper`.** Jackson already knows how to turn a `String`/`Number`/`Map`/`List` into any registered type via `ObjectMapper.convertValue(raw, JavaType)` — the runtime just needs to stop gating that call on `Map`/`List`.

### Runtime (`FeatureManagerBase.handleParam`)

1. Skip the `isValid(val)` structural check for target types that are not one of the trivial runtime-handled types (primitive wrappers, `String`, enum, `Map`, `List`, array). Let Jackson decide.
2. Collapse the `Map` / `List` / fallback branches into a single call: if `val.getClass()` is not assignable to the target type, do `mapper.convertValue(val, mapper.getTypeFactory().constructType(arg.type()))`. That single call covers:
   - `Map` → POJO (current behavior)
   - `List` → `List<T>`/array (current behavior)
   - `String` → custom type with a registered `JsonDeserializer` or `@JsonCreator(String)` (currently broken)
   - `Number` → custom type with a numeric deserializer (currently broken)
   - Enum lookup can also be delegated to Jackson, removing the bespoke `Enum.valueOf` branch (and getting `@JsonCreator`/`@JsonValue`/`@JsonAlias` for free).

Sketch:

```java
private Object handleParam(FeatureMetadata<?> metadata, String serverName, FeatureArgument arg, Object val) {
    if (val == null) {
        if (arg.defaultValue() != null) val = convert(arg.defaultValue(), arg.type());
        if (val == null && arg.required()) throw invalidArgument(metadata, serverName, "Missing required argument: " + arg.name());
    } else {
        JavaType javaType = mapper.getTypeFactory().constructType(arg.type());
        if (!javaType.getRawClass().isInstance(val)) {
            try {
                val = mapper.convertValue(val, javaType);
            } catch (IllegalArgumentException e) {
                throw invalidArgument(metadata, serverName,
                        "Invalid argument [%s] - unable to convert to %s: %s"
                                .formatted(arg.name(), javaType, e.getOriginalMessage()));
            }
        }
    }
    return arg.isOptional() ? Optional.ofNullable(val) : val;
}
```

### Schema (`DefaultSchemaGenerator`)

Drive the schema off the `ObjectMapper` for types the built-in enumeration doesn't recognize, so the schema matches what Jackson actually emits/accepts. Two options:

- **Preferred:** use `ObjectMapper.acceptJsonFormatVisitor(JavaType, visitor)` on a [`JsonFormatVisitorWrapper`](https://javadoc.io/doc/com.fasterxml.jackson.core/jackson-databind/latest/com/fasterxml/jackson/databind/jsonFormatVisitors/JsonFormatVisitorWrapper.html) to resolve the real serialized JSON type (including types with a `@JsonSerialize`/custom serializer that narrows to string/number/etc.). Fall back to `victools` only when the visitor declines. This gives `MyDto -> string` without any user configuration.
- **Alternative:** let users register a `SchemaGeneratorConfigCustomizer` or a new `@ToolArg`-level `schemaType` override. Less magical, but requires explicit declaration for every custom scalar and doesn't fix the default experience.

Alongside that, `FeatureArgument.expectedJsonTypeFrom` should consult the same visitor instead of defaulting unknown types to `OBJECT`, so runtime validation and the generated schema agree. This removes the need for the special branch in `handleParam` to coerce string/number into custom types via `if/else` chains — both paths end up asking Jackson the same question.

## Why `ObjectMapper.convertValue`

- It already respects every user-registered `Module`, `JsonDeserializer`, `@JsonCreator`, `@JsonValue`, and mix-in — which is exactly the extensibility contract users expect when they configure a CDI-produced `ObjectMapper` for their Quarkus app.
- It's idempotent when the source already matches the target, so "same type → pass through" falls out naturally (`javaType.getRawClass().isInstance(val)` short-circuits).
- It keeps a single code path instead of the current `Map`/`List`/enum/number ladder, which is the source of the asymmetry.

## Workaround today

Declare custom-typed args as `String` and parse them inside the tool (throwing `ToolCallException` on parse failure). This doesn't feel like the same flow as when using quarkus rest and isn't intuitive.

Additionally, I'm not sure if it's intentional, but the ObjectMapper used to generate the JSON schema is not the Quarkus CDI one,
causing some of the issues above (serializers not used). It seems to make sense that the one used at runtime (CDI) should match
the one used to generate the schema (created by the constructor inside jsonschema-generator).

Happy to work on a PR for this if you agree with the direction (essentially delegating a lot more to the ObjectMapper and using the CDI ObjectMapper to generate schemas as well as at runtime).

mkouba:
- fixes #774
- remove validation, always use Jackson for conversion to support custom serializers/deserializers
- also add custom serializer input schema generator test

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/DefaultSchemaGenerator.java
+++ b/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/DefaultSchemaGenerator.java
@@ -7,6 +7,7 @@ import java.util.Map;
 import jakarta.inject.Singleton;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.github.victools.jsonschema.generator.Option;
 import com.github.victools.jsonschema.generator.OptionPreset;
@@ -35,8 +36,9 @@ public class DefaultSchemaGenerator implements GlobalInputSchemaGenerator, Globa
     final Map<Type, DefaultValueConverter<?>> defaultValueConverters;
 
     public DefaultSchemaGenerator(@All List<SchemaGeneratorConfigCustomizer> schemaGeneratorConfigCustomizers,
+            ObjectMapper objectMapper,
             McpMetadata metadata) {
-        this.schemaGenerator = constructSchemaGenerator(schemaGeneratorConfigCustomizers);
+        this.schemaGenerator = constructSchemaGenerator(objectMapper, schemaGeneratorConfigCustomizers);
         this.toolArgumentHolders = metadata.toolArgumentHolders();
         this.defaultValueConverters = metadata.defaultValueConverters();
     }
@@ -102,8 +104,9 @@ public class DefaultSchemaGenerator implements GlobalInputSchemaGenerator, Globa
     }
 
     private static SchemaGenerator constructSchemaGenerator(
+            ObjectMapper objectMapper,
             List<SchemaGeneratorConfigCustomizer> schemaGeneratorConfigCustomizers) {
-        var configBuilder = new SchemaGeneratorConfigBuilder(SchemaVersion.DRAFT_2020_12, OptionPreset.PLAIN_JSON)
+        var configBuilder = new SchemaGeneratorConfigBuilder(objectMapper, SchemaVersion.DRAFT_2020_12, OptionPreset.PLAIN_JSON)
                 .without(Option.SCHEMA_VERSION_INDICATOR);
         for (SchemaGeneratorConfigCustomizer customizer : schemaGeneratorConfigCustomizers) {
             customizer.customize(configBuilder);
--- a/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/FeatureArgument.java
+++ b/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/FeatureArgument.java
@@ -1,8 +1,6 @@
 package io.quarkiverse.mcp.server.runtime;
 
 import java.lang.reflect.ParameterizedType;
-import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 import java.util.OptionalDouble;
 import java.util.OptionalInt;
@@ -17,7 +15,6 @@ public record FeatureArgument(String name,
         java.lang.reflect.Type type,
         String defaultValue,
         Provider provider,
-        JsonType expectedJsonType,
         OptionalKind optionalKind) {
 
     // this constructor is used for McpMetadata
@@ -28,16 +25,13 @@ public record FeatureArgument(String name,
             java.lang.reflect.Type type,
             String defaultValue,
             Provider provider) {
-        this(name, title, description, required, type, defaultValue, provider, null, OptionalKind.NONE);
+        this(name, title, description, required, type, defaultValue, provider, OptionalKind.NONE);
     }
 
     public FeatureArgument {
         if (Types.isOptional(type)) {
             optionalKind = OptionalKind.of(type);
             type = Types.getOptionalValueType(type);
-        }
-        if (expectedJsonType == null) {
-            expectedJsonType = expectedJsonTypeFrom(type);
         }
     }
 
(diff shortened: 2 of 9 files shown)


mkouba:
CC @jameswilliams1

jameswilliams1 (commented):
thanks @mkouba ! this looks like it would work to me (the custom serializer not working for schema gen is probably a separate issue, and easy to work around with jackson annotations anyway). Guess this needs to be reviewed by quarkus team but very keen to test on my side once a tag is ready

mkouba:
> thanks @mkouba ! this looks like it would work to me (the custom serializer not working for schema gen is probably a separate issue, and easy to work around with jackson annotations anyway). 

Yes, it's a separate issue and I'm afraid that we can't fix it reliably anyway.

> Guess this needs to be reviewed by quarkus team but very keen to test on my side once a tag is ready

Yes, I checked the code and it looks good, so I will "self-approve" ;-).

PR #780, opened 2026-05-08
Add an extension to deal with OIDC specific MCP authorization requirements

sberyozkin:
To address a `SHOULD` recommendation at https://modelcontextprotocol.io/specification/draft/basic/authorization#runtime-insufficient-scope-errors, where `403` is accompanied by `WWW-Authenticate` with a list of expected scopes and a link to the OAuth2 resource metadata handler (that would typically be done only for `401`), an OIDC specific `ForbiddenException` handler has to be provided. This handler can't really be shipped directly with the `transports/http`.

Therefore, a `quarkus-mcp-server-oidc` extension is proposed, similarly to how it is done with the hibernate validator.

Please note that even though this extension only holds a Vert.x failure handler for now, we can definitely make it more feature-rich going forward.

For example, one thing I've been thinking about is, instead of asking users to add OIDC configuration manually, this extension might be able to listen to Streambale MCP route registration events and register an OIDC handler automatically.

I hope we will be able to do something interesting in this extension in the future.

### DIFF
--- a/transports/http/authorization-integration-tests/pom.xml
+++ b/transports/http/authorization-integration-tests/pom.xml
@@ -45,6 +45,24 @@
             </exclusions>
         </dependency>
         <dependency>
+            <groupId>io.quarkiverse.mcp</groupId>
+            <artifactId>quarkus-mcp-server-oidc</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkiverse.mcp</groupId>
+            <artifactId>quarkus-mcp-server-oidc-deployment</artifactId>
+            <version>${project.version}</version>
+            <type>pom</type>
+            <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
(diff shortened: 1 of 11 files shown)


transports/http/authorization-integration-tests/pom.xml
```diff
@@ -30,6 +30,11 @@
             <artifactId>quarkus-mcp-server-http</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>io.quarkiverse.mcp</groupId>
+            <artifactId>quarkus-mcp-server-oidc</artifactId>
```
mkouba, line 49: @sberyozkin I think that you also need to add the `quarkus-mcp-server-oidc-deployment` dependency next to the [`quarkus-mcp-server-http-deployment`](https://github.com/quarkiverse/quarkus-mcp-server/blob/main/transports/http/integration-tests/pom.xml#L36-L49) for consistent build order.

mkouba (approved):

PR #812, opened 2026-06-02
core: add Optional<Method> method() to FeatureInfo

### LINKED ISSUE #809: Expose declaring class in e.g. io.quarkiverse.mcp.server.ToolFilter chain
My own auth logic wants a custom annotation with @com.myproject.Secured information on the Tool class itself. But when the filter chain runs, I cannot get a hold of the actual tool class.

If I could get it from io.quarkiverse.mcp.server.ToolManager.ToolInfo, my implementation would be very smooth. I would just check the class level annotation on that class, compare against the auth response and be done.

Now I have to have static names for all tools and as a side step scan for all tool classes and make a mapping.

mkouba:
- resolves #809


mkouba:
FYI @karlkilden

PR #819, opened 2026-06-09
Avoid NPE for guarded no-argument tool calls

moayad-alyaghshi:
This fixes a `NullPointerException` when a no-argument tool has an input guardrail and a client omits the `arguments` field from a `tools/call` request. I first reproduced this with `McpStreamableTestClient`, when it sends no `arguments` field for no-argument tool calls.

Input guardrails are not only used for argument validation. They can also be used for cross-cutting pre-call checks such as rate limiting. For example, a no-argument read tool may still need a rate-limit guardrail even though it has no arguments to validate.

Without this fix, attaching any input guardrail to a no-argument tool is unsafe for clients that omit `arguments`. The framework throws an NPE while building the guardrail input context, before the guardrail logic can run. As a result, the request fails with an internal error instead of being handled by the guardrail or the tool.

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/ToolManagerImpl.java
+++ b/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/ToolManagerImpl.java
@@ -439,6 +439,9 @@ public class ToolManagerImpl extends FeatureManagerBase<ToolResponse, ToolInfo> 
 
         @Override
         public void setArguments(JsonObject arguments) {
+            if (arguments == null) {
+                arguments = new JsonObject();
+            }
             JsonObject newArgs = new JsonObject(Map.copyOf(arguments.getMap()));
             this.arguments.set(newArgs);
         }


mkouba:
@moayad-alyaghshi Good catch! However, I think we should disallow passing `null`, i.e. just use `Objects.requireNonNull()` and clarify the javadoc of `ToolInputGuardrail.ToolInputContext.setArguments()`.

mkouba:
It would be also great to have an automated test like [this](https://github.com/quarkiverse/quarkus-mcp-server/pull/820/changes#diff-841c9fb6936ae2b967861e486530147f87ce28fbbd437221652e806ab30b82c0R21-R63) :wink:.

mkouba:
@moayad-alyaghshi if you don't have time to update this PR, I can take it over.

moayad-alyaghshi:
@mkouba I can work on it tomorrow morning. If you prefer to take care of it before then, that’s completely fine with me too. Thanks.

mkouba:
Superseded by https://github.com/quarkiverse/quarkus-mcp-server/pull/831.

moayad-alyaghshi:
@mkouba https://github.com/quarkiverse/quarkus-mcp-server/pull/831 hardens `setArguments(null)`, but it does not address the original failure path. The crash I saw was not from a guardrail explicitly calling `context.setArguments(null)`. It happened while the framework was constructing `ToolInputContextImpl`:

```java
setArguments(Messages.getArguments(...))
```

When a no-argument tool call omits the `arguments` field, `Messages.getArguments(...)` returns `null`. With this change, that path will now fail on `Objects.requireNonNull(arguments)` instead of `arguments.getMap()`, but the guarded no-argument tool call still fails before the guardrail can run.

I think the constructor needs to normalize the request value before calling `setArguments(...)`, while keeping `setArguments(null)` invalid as an API contract. For example:

```java
JsonObject arguments = Messages.getArguments(getParams(message));
setArguments(arguments != null ? arguments : new JsonObject());
```

It would also be useful for the regression test to cover the original request shape: a guarded no-argument tool called without `params.arguments`. The current test covers explicit `setArguments(null)` from inside a guardrail, which is a different path.

PR #831, opened 2026-06-10
ToolInputGuardrail.ToolInputContext#setArguments() should not accept null

mkouba:
@moayad-alyaghshi thanks!

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/mcp/server/ToolInputGuardrail.java
+++ b/core/runtime/src/main/java/io/quarkiverse/mcp/server/ToolInputGuardrail.java
@@ -90,7 +90,7 @@ public interface ToolInputGuardrail {
         /**
          * Set the current arguments.
          *
-         * @param arguments
+         * @param arguments must not be {@code null}
          */
         void setArguments(JsonObject arguments);
 
--- a/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/ToolManagerImpl.java
+++ b/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/ToolManagerImpl.java
@@ -430,7 +430,8 @@ public class ToolManagerImpl extends FeatureManagerBase<ToolResponse, ToolInfo> 
         ToolInputContextImpl(ToolInfo tool, JsonObject message, McpRequest request) {
             super(tool, message, request);
             this.arguments = new AtomicReference<>();
-            setArguments(Messages.getArguments(getParams(message)));
+            JsonObject args = Messages.getArguments(getParams(message));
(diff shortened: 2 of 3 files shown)


moayad-alyaghshi:
@mkouba This hardens `setArguments(null)`, but it does not address the original failure path. The crash I saw was not from a guardrail explicitly calling `context.setArguments(null)`. It happened while the framework was constructing `ToolInputContextImpl`:

```java
setArguments(Messages.getArguments(...))
```

When a no-argument tool call omits the `arguments` field, `Messages.getArguments(...)` returns `null`. With this change, that path will now fail on `Objects.requireNonNull(arguments)` instead of `arguments.getMap()`, but the guarded no-argument tool call still fails before the guardrail can run.

I think the constructor needs to normalize the request value before calling `setArguments(...)`, while keeping `setArguments(null)` invalid as an API contract. For example:

```java
JsonObject arguments = Messages.getArguments(getParams(message));
setArguments(arguments != null ? arguments : new JsonObject());
```

It would also be useful for the regression test to cover the original request shape: a guarded no-argument tool called without `params.arguments`. The current test covers explicit `setArguments(null)` from inside a guardrail, which is a different path.

mkouba:
> @mkouba This hardens `setArguments(null)`, but it does not address the original failure path. The crash I saw was not from a guardrail explicitly calling `context.setArguments(null)`. It happened while the framework was constructing `ToolInputContextImpl`:
> 
> ```java
> setArguments(Messages.getArguments(...))
> ```
> 
> When a no-argument tool call omits the `arguments` field, `Messages.getArguments(...)` returns `null`. With this change, that path will now fail on `Objects.requireNonNull(arguments)` instead of `arguments.getMap()`, but the guarded no-argument tool call still fails before the guardrail can run.
> 
> I think the constructor needs to normalize the request value before calling `setArguments(...)`, while keeping `setArguments(null)` invalid as an API contract. For example:
> 
> ```java
> JsonObject arguments = Messages.getArguments(getParams(message));
> setArguments(arguments != null ? arguments : new JsonObject());
> ```
> 
> It would also be useful for the regression test to cover the original request shape: a guarded no-argument tool called without `params.arguments`. The current test covers explicit `setArguments(null)` from inside a guardrail, which is a different path.

I see. You're absolutely right and I will update the fix.

mkouba:
@moayad-alyaghshi fixed!

mkouba:
I'm going to merge (and backport) this pull request. If we find another problem, we'll address that in a follow-up.

moayad-alyaghshi:
@mkouba thank you

mkouba:
> @mkouba thank you

Thank you for the contribution!

PR #850, opened 2026-06-17
Tracing spans include mcp.server.name and respect per-server config

### LINKED ISSUE #845: Tracing spans do not include the name of MCP server configuration
All MCP spans should include the `mcp.server.name` attribute.

Moreover, we should not start a span if `quarkus.mcp.server.tracing.enabled=false` for the relevant MCP server config.

mkouba:
- all MCP tracing spans now include the mcp.server.name attribute
- per-server tracingEnabled check is performed at span start time instead of only checking the default server at construction time
- resolves #845

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/tracing/McpTracingInstrumenter.java
+++ b/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/tracing/McpTracingInstrumenter.java
@@ -29,7 +29,6 @@ import io.opentelemetry.semconv.incubating.GenAiIncubatingAttributes;
 import io.opentelemetry.semconv.incubating.RpcIncubatingAttributes;
 import io.quarkiverse.mcp.server.InitialRequest;
 import io.quarkiverse.mcp.server.McpMethod;
-import io.quarkiverse.mcp.server.McpServer;
 import io.quarkiverse.mcp.server.runtime.McpRequest;
 import io.quarkiverse.mcp.server.runtime.McpTracing;
 import io.quarkiverse.mcp.server.runtime.McpTracingSpan;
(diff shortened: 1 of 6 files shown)


core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/tracing/McpTracingInstrumenter.java
```diff
@@ -79,6 +80,9 @@ public McpTracingInstrumenter(OpenTelemetry openTelemetry, OTelRuntimeConfig ote
     @Override
     public McpTracingSpan startSpan(McpMethod method, JsonObject message, McpRequest mcpRequest,
             InitialRequest.Transport transport) {
+        if (!mcpConfig.servers().get(mcpRequest.serverName()).tracingEnabled()) {
```
brunobat: If `otelConfig.sdkDisabled()` It should also return a NOOP.

mkouba: Hm, we already use `otelConfig.sdkDisabled()` for `InstrumenterBuilder.setEnabled()` in the constructor. Isn't it enough?

brunobat: The instrumenter will not produce any work, however, the `public McpTracingSpan startSpan` method could probably return sooner. If no instrumentation is on, no need to execute all the logic in the method.

mkouba: Done.

Issue #788, opened 2026-05-15
The type of Annotations.audience should be Role[], not Role


pweicken:
The MCP specification defines the audience of the Annotations type as a Role[]. 
(https://modelcontextprotocol.io/specification/2025-11-25/schema#annotations)

However, the implementation in the io.quarkiverse.mcp.server.Content.Annotations record defines it as a single Role.
This causes the output of the MCP server to be wrong, i.e.
Expected: `audience: ["user"]`
Actual: `audience: "user"`

Workaround for single values:
Its possible to output an array by adding a custom JsonSerializer for Content.Annotations.

mkouba:
Thanks for the report. This is obviously an oversight. Unfortunately, the fix will be a breaking change.

mkouba:
FYI I've just released [1.12.1](https://github.com/quarkiverse/quarkus-mcp-server/releases/tag/1.12.1) that includes the fix for this issue.

pweicken:
Thank you for the quick fix!

PR #853, opened 2026-06-17
fix: always end the response in the auth failure handler

SaSteffen:
## Problem

`createAuthFailureHandler` (added in #788) intercepts auth rejections at
the MCP endpoint and tries to extract a JSON-RPC `id` from the request
body so it can send a well-formed JSON-RPC error response.

The handler only calls `ctx.response().end()` inside the success branch:

```java
ctx.request().body().onComplete(ar -> {
    if (ar.succeeded() && ar.result() != null) {
        Integer id = new JsonObject(ar.result()).getInteger("id"); // throws on empty body
        ...
        ctx.response().putHeader(...).end(error.toBuffer());
    }
    // ← nothing here: response is never ended if body is empty
});
```

For **GET requests** — which Streamable HTTP clients use to open an SSE
stream — the body is empty. `new JsonObject(emptyBuffer)` throws a
`DecodeException` before `end()` is reached. The response status is set
to 401 but the connection is never closed; the client hangs instead of
receiving the rejection.

The same defect fires when the body future itself fails (e.g. body
already consumed by the time the failure route runs).

## Fix

Move the JSON-RPC error assembly and `end()` call outside the condition
so they always execute. Extract the `id` with a try/catch; on failure
(empty or non-JSON body) `id` stays `null`, which is valid per the
JSON-RPC spec for requests whose `id` cannot be determined.

```java
ctx.request().body().onComplete(ar -> {
    Integer id = null;
    if (ar.succeeded() && ar.result() != null && !ar.result().isEmpty()) {
        try {
            id = new JsonObject(ar.result()).getInteger("id");
        } catch (Exception ignored) {
        }
    }
    JsonObject error = new JsonObject()
            .put("jsonrpc", "2.0")
            .put("id", id)
            .put("error", ...);
    ctx.response().putHeader(...).end(error.toBuffer()); // always reached
});
```

## Impact

Before this fix, any MCP client that sends a GET `/mcp` request with an
invalid or expired token (to reconnect an SSE stream) receives a
hanging connection rather than a clean 401. Observed against
`quarkus-mcp-server` 1.13.0.

### DIFF
--- a/transports/http/runtime/src/main/java/io/quarkiverse/mcp/server/http/runtime/HttpMcpServerRecorder.java
+++ b/transports/http/runtime/src/main/java/io/quarkiverse/mcp/server/http/runtime/HttpMcpServerRecorder.java
@@ -121,16 +121,21 @@ public class HttpMcpServerRecorder {
                         ctx.request().resume();
                     }
                     ctx.request().body().onComplete(ar -> {
-                        if (ar.succeeded() && ar.result() != null) {
-                            Integer id = new JsonObject(ar.result()).getInteger("id");
-                            JsonObject error = new JsonObject()
-                                    .put("jsonrpc", "2.0")
-                                    .put("id", id)
-                                    .put("error", new JsonObject()
-                                            .put("code", JsonRpcErrorCodes.SECURITY_ERROR)
-                                            .put("message", failure.toString()));
-                            ctx.response().putHeader(HttpHeaders.CONTENT_TYPE, "application/json").end(error.toBuffer());
+                        Object id = null;
+                        if (ar.succeeded() && ar.result() != null && ar.result().length() > 0) {
+                            try {
+                                id = new JsonObject(ar.result()).getValue("id");
+                            } catch (Exception e) {
+                                LOG.debugf(e, "Unable to parse request body");
+                            }
                         }
+                        JsonObject error = new JsonObject()
+                                .put("jsonrpc", "2.0")
+                                .put("id", id)
+                                .put("error", new JsonObject()
+                                        .put("code", JsonRpcErrorCodes.SECURITY_ERROR)
+                                        .put("message", failure.toString()));
+                        ctx.response().putHeader(HttpHeaders.CONTENT_TYPE, "application/json").end(error.toBuffer());
                     });
                 } else {
                     ctx.next();


transports/http/runtime/src/main/java/io/quarkiverse/mcp/server/http/runtime/HttpMcpServerRecorder.java
```diff
@@ -121,16 +121,21 @@ public void handle(RoutingContext ctx) {
...
-                            Integer id = new JsonObject(ar.result()).getInteger("id");
-                            JsonObject error = new JsonObject()
-                                    .put("jsonrpc", "2.0")
-                                    .put("id", id)
-                                    .put("error", new JsonObject()
-                                            .put("code", JsonRpcErrorCodes.SECURITY_ERROR)
-                                            .put("message", failure.toString()));
-                            ctx.response().putHeader(HttpHeaders.CONTENT_TYPE, "application/json").end(error.toBuffer());
+                        Integer id = null;
+                        if (ar.succeeded() && ar.result() != null && !ar.result().isEmpty()) {
```
mkouba: `io.vertx.core.buffer.Buffer` does not declare the method `isEmpty()`. Did you even try to compile your code?

transports/http/runtime/src/main/java/io/quarkiverse/mcp/server/http/runtime/HttpMcpServerRecorder.java
```diff
@@ -121,16 +121,21 @@ public void handle(RoutingContext ctx) {
...
+                            try {
+                                id = new JsonObject(ar.result()).getInteger("id");
+                            } catch (Exception ignored) {
+                                // Body is not valid JSON (e.g. GET request with empty body)
```
mkouba: I think that we should log a warning here.

transports/http/runtime/src/main/java/io/quarkiverse/mcp/server/http/runtime/HttpMcpServerRecorder.java
```diff
@@ -121,16 +121,21 @@ public void handle(RoutingContext ctx) {
...
-                        if (ar.succeeded() && ar.result() != null) {
-                            Integer id = new JsonObject(ar.result()).getInteger("id");
-                            JsonObject error = new JsonObject()
-                                    .put("jsonrpc", "2.0")
-                                    .put("id", id)
-                                    .put("error", new JsonObject()
-                                            .put("code", JsonRpcErrorCodes.SECURITY_ERROR)
-                                            .put("message", failure.toString()));
-                            ctx.response().putHeader(HttpHeaders.CONTENT_TYPE, "application/json").end(error.toBuffer());
+                        Integer id = null;
```
mkouba: `Integer` is actually wrong, because in MCP a request id can be either string or int. So we should use `Object` here.

This is a different issue but we could fix it here.

mkouba (changes requested):
Thanks for the pull request. I've added a few comments.

CC @sberyozkin

transports/http/runtime/src/main/java/io/quarkiverse/mcp/server/http/runtime/HttpMcpServerRecorder.java
SaSteffen: Sorry, to bother you, didn't mean to open this in ready-state. Set as draft

transports/http/runtime/src/main/java/io/quarkiverse/mcp/server/http/runtime/HttpMcpServerRecorder.java
SaSteffen: Added a debug message, so malformed bodies don't flood server logs. This log-level would be in line with other client-side error logs in the file.

sberyozkin (approved):
Thanks @SaSteffen @mkouba

mkouba (approved):

mkouba:
Backport PR: https://github.com/quarkiverse/quarkus-mcp-server/pull/855

PR #864, opened 2026-06-23
Add McpTrafficListener API for custom traffic logging

### LINKED ISSUE #860: Provide an API to add custom TrafficLogger
Today TrafficLogger is used to show MCP request/response in logs.

In my case I would like to provide in my MCP Server an Admin which shows in a MCP console (per AI client) those logs (with folding etc) like I did for LSP:

Image

mkouba:
- resolves #860


mkouba:
CC @angelozerr

angelozerr:
@mkouba thanks for your PR!

In my case I would like to track messages but without logging message in console.

One idea that I have is to move 

`LOG.infof("MCP message received [%s]:\n\n%s", connection.id(), messageToString(message, textPayloadLimit));`

in a class `DefaultMcpTrafficListener` and have a quarkus property to disable it (enable by default).

I wonder also if `RawMessage #asString()` could cache the encoded String.

If there are N McpTrafficListener their will hav ejust one call of encode method.

mkouba:
> @mkouba thanks for your PR!
> 
> In my case I would like to track messages but without logging message in console.
> 
> One idea that I have is to move
> 
> `LOG.infof("MCP message received [%s]:\n\n%s", connection.id(), messageToString(message, textPayloadLimit));`
> 
> in a class `DefaultMcpTrafficListener` and have a quarkus property to disable it (enable by default).

I see. In that case, we would need to make the `TrafficLogger` implement the `McpTrafficListener` and call the listeners even if `quarkus.mcp-server.traffic-logging.enabled=false`, which is suboptimal in terms of performance but probably an acceptable overhead.

> I wonder also if `RawMessage #asString()` could cache the encoded String.
> 
> If there are N McpTrafficListener their will hav ejust one call of encode method.

Yes, `asString()` could be lazily cached. `asJsonObject()`  unfortunately not.

I will try to update the PR later today.

mkouba:
@angelozerr PR updated, all comments should be addressed now.

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/RawMessageImpl.java
angelozerr, line 26: You return a copy because you want to be sure that jdon object cannot be updated by listener. Do you think really that listener could update it?

If you want to really keeps that, it would be very nice  if RawMessage could have Integer getId()

This id is very helpfull to generate time like lsp traces does.

It is out of the scope of this issue, but I think quarkus mcp server should provides that with config (simple like today and withTime)

angelozerr (approved):

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/RawMessageImpl.java
mkouba: > Do you think really that listener could update it?

Yes, it can. A listener could just call `JsonObject#put()` :shrug:.

> If you want to really keeps that, it would be very nice if RawMessage could have Integer getId()

Yes, that would be helpful, except that (a) the id is not always present, and (b) it could also be a string per the spec. I'll try to add something.

> It is out of the scope of this issue, but I think quarkus mcp server should provides that with config (simple like today and withTime)

I don't understand. Could you be more specific? @angelozerr

angelozerr: > > Do you think really that listener could update it?
> 
> Yes, it can. A listener could just call `JsonObject#put()` 🤷.<

Yes sure but do you think really that a listener wants to do that?

> 
> > If you want to really keeps that, it would be very nice if RawMessage could have Integer getId()
> 
> Yes, that would be helpful, except that (a) the id is not always present, and (b) it could also be a string per the spec. I'll try to add something.

LSP has the same issues (id=Integer|string). LSP4J uses https://github.com/eclipse-lsp4j/lsp4j/blob/main/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/messages/Either.java

You could have `Either<Integer,String> getId()` which can return null.

> 
> > It is out of the scope of this issue, but I think quarkus mcp server should provides that with config (simple like today and withTime)
> 
> I don't understand. Could you be more specific? @angelozerr

See https://github.com/quarkiverse/quarkus-mcp-server/issues/860#issuecomment-4776835126

```
[Trace - 00:46:53] Sending request 'initialize - (0)'
[Trace - 00:46:53] Received response 'initialize - (0)' in 15ms
[Trace - 00:46:53] Received notification 'notifications/initialized'
[Trace - 00:46:53] Sending notification 'notifications/message'
[Trace - 00:46:54] Sending request 'tools/call - (1)'
[Trace - 00:47:02] Received response 'tools/call - (1)' in 8821ms
[Trace - 00:47:09] Sending request 'tools/call - (2)'
[Trace - 00:47:09] Received response 'tools/call - (2)' in 26ms
[Trace - 00:47:14] Sending request 'tools/call - (3)'
[Trace - 00:47:14] Received response 'tools/call - (3)' in 30ms
[Trace - 00:47:19] Sending request 'tools/call - (4)'
[Trace - 00:47:19] Received response 'tools/call - (4)' in 30ms
[Trace - 00:51:45] Sending notification 'notifications/message'
[Trace - 00:56:46] Sending notification 'notifications/message'
[Trace - 00:57:07] Sending request 'tools/call - (5)'
[Trace - 00:57:07] Received response 'tools/call - (5)' in 33ms
[Trace - 00:57:11] Sending request 'tools/call - (6)'
[Trace - 00:57:11] Received response 'tools/call - (6)' in 70ms
[Trace - 00:57:53] Sending request 'tools/call - (7)'
[Trace - 00:57:53] Received response 'tools/call - (7)' in 24ms
[Trace - 00:58:01] Sending request 'tools/call - (8)'
[Trace - 00:58:01] Received response 'tools/call - (8)' in 25ms
[Trace - 01:01:47] Sending notification 'notifications/message'
[Trace - 01:06:47] Sending notification 'notifications/message'
[Trace - 01:21:48] Sending request 'tools/call - (9)'
[Trace - 01:21:48] Received response 'tools/call - (9)' in 33ms
[Trace - 01:30:24] Sending notification 'notifications/message'
[Trace - 02:31:51] Sending notification 'notifications/message'
[Trace - 02:36:53] Sending notification 'notifications/message'
[Trace - 02:41:53] Sending notification 'notifications/message'
```

mkouba: > > > Do you think really that listener could update it?
> > 
> > 
> > Yes, it can. A listener could just call `JsonObject#put()` 🤷.<
> 
> Yes sure but do you think really that a listener wants to do that?

It's not a question whether a listener _wants to do that_, but whether it _can do that_. If it can, we have to take it into account.

> 
> > > If you want to really keeps that, it would be very nice if RawMessage could have Integer getId()
> > 
> > 
> > Yes, that would be helpful, except that (a) the id is not always present, and (b) it could also be a string per the spec. I'll try to add something.
> 
> LSP has the same issues (id=Integer|string). LSP4J uses https://github.com/eclipse-lsp4j/lsp4j/blob/main/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/messages/Either.java
> 
> You could have `Either<Integer,String> getId()` which can return null.

Yes, I was thinking about `Object getId()` which returns `String`, `Number`, or `null`.

> 
> > > It is out of the scope of this issue, but I think quarkus mcp server should provides that with config (simple like today and withTime)
> > 
> > 
> > I don't understand. Could you be more specific? @angelozerr
> 
> See [#860 (comment)](https://github.com/quarkiverse/quarkus-mcp-server/issues/860#issuecomment-4776835126)
> 
> ```
> [Trace - 00:46:53] Sending request 'initialize - (0)'
> ```

Ok, that's really out of scope.

angelozerr: > Yes, I was thinking about Object getId() which returns String, Number, or null.

IMHO Either should be better ad consistent with LSP4J

mkouba: > > Yes, I was thinking about Object getId() which returns String, Number, or null.
> 
> IMHO Either should be better ad consistent with LSP4J

Except that we don't have LSP4J on the class path and we will not add that dependency to MCP server.

Moreover, it's IMO an _overkill_ in this particular case.

angelozerr: I though that you copy/paste Either in Quarkus MCP Server, but what about existing RequestId record ?

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/RawMessageImpl.java
```diff
@@ -23,7 +25,17 @@ public JsonObject asJsonObject() {
...
+        String s = cachedString;
+        if (s == null) {
+            s = message.encode();
+            cachedString = s;
+        }
+        return s;
+    }
+
+    @Override
+    public Object id() {
```
angelozerr: What about having `RequestId getId()`

mkouba: Well, `RequestId` would be more consitent. But TBH I don't think it's a great API. Will try that later today.

mkouba: Done.

angelozerr: Thanks! Just one question, we can use RequestId as key in a Map right?

angelozerr:
@mkouba as we have now `RawMessage#getId()` it would be very nice to have also `RawMessage#getMethod()` which can return null.

In my case I need it to generate proper log header message and it will avoid to use JsonObject (do a copy)

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/RawMessageImpl.java
mkouba: > Thanks! Just one question, we can use RequestId as key in a Map right?

Yes, since the value is either `Number` or `String` and `RequestId` has equals/hashCode based on the value.

angelozerr: That's great, I will able to use this request id to manage pending request to generate time in the header log (like LSP trace)

mkouba:
> @mkouba as we have now `RawMessage#getId()` it would be very nice to have also `RawMessage#getMethod()` which can return null.
> 
> In my case I need it to generate proper log header message and it will avoid to use JsonObject (do a copy)

Good idea, added.

@angelozerr Let me know if you have some other ideas for improvements. We could address them in a separate PR though. I'd like to merge this one today.

angelozerr:
@mkouba I think your PR will cover all my need to generate proper log with timing.

I have not tested but it looks really very good!

Thanks so much @mkouba !

PR #867, opened 2026-06-25
Cancellation#onCancelled() callback for push-based notifications

### LINKED ISSUE #866: Provide capability to subscribe to cancel request
I have a tool which looks like this:

```java
@Tool
String findReferences(String uri, int line, int character) {
  // 1. Get future which starts LS server
  CompletableFuture loadServer = ....
  return loadServer
     .theCompose((loadedServer) -> {
         // 2. Get future from LSP textDocument/refeences
         CompletableFuture loadReferences = loadedServer.getTextDocumentService().references(.....)
         return loadReferences;
     });
}
```

Now I want to cancel the 2) future when for instance user will cancel (with ESC) the MCP client

To do that I would like to use Cancellation cancellation like this:

  ```java
cancellation.subscribe(new io.smallrye.mutiny.subscription.Cancellable() {
       void cancel() {
           loadReferences.cancel(true);
       }
 }
```

```java
@Tool
String findReferences(String uri, int line, int character, Cancellation cancellation) {
  // 1. Get future which starts LS server
  CompletableFuture loadServer = ....
  return loadServer
     .theCompose((loadedServer) -> {
         // 2. Get future from LSP textDocument/refeences
         CompletableFuture loadReferences = loadedServer.getTextDocumentService().references(.....)
         

         // HERE I WANT TO SUBSCRIBE TO cancellation. When cancellation will be cancelled, it should cancel the loadReferences future:
        cancellation.subscribe(new io.smallrye.mutiny.subscription.Cancellable() {
             void cancel() {
                 loadReferences.cancel(true);
             }
       }

         return loadReferences;
     });
}
```

mkouba:
- register an action that is executed when a cancellation of the current MCP request is requested by the client, instead of polling with check()
- resolves #866

CC @angelozerr


core/runtime/src/main/java/io/quarkiverse/mcp/server/Cancellation.java
```diff
@@ -24,6 +24,15 @@ public interface Cancellation {
...
     Result check();
 
+    /**
+     * Registers an action that is executed when a cancellation of the current MCP request is requested by the client. If
+     * cancellation was already requested, the action is executed immediately. Multiple actions can be registered.
+     *
+     * @param action the action to execute on cancellation, must not be {@code null}
+     * @throws IllegalArgumentException if the action is {@code null}
+     */
+    void onCancelled(Runnable action);
```
angelozerr: I wonder if instead of having Runnable, we could have Consumer<String> where string is the cancel reason?

mkouba: Yes, that's a good idea, although the reason is optional. Hm, `Consumer<Optional<String>>` looks awkward. But a dedicated functional interface would not be any better...

mkouba: Done.

angelozerr:
@mkouba I have not tested but API looks very good. Thanks!

PR #878, opened 2026-07-01
Handle notifications/initialized on auto-init path

mkouba:
- when notifications/initialized arrives with an unknown session id and
 auto-init is enabled, the notification now returns 202 instead of an
 unsupported method error


mkouba:
The [98469da](https://github.com/quarkiverse/quarkus-mcp-server/pull/878/commits/98469da8bb78c986c261c015953e915c788d42f0) commit could be also backported to 1.13.x.

ikstewa (approved):

PR #895, opened 2026-07-09
Fix output schema generation for generic return types

### LINKED ISSUE #894: Tools with structuredContent do not support parameterized return types
When a `@Tool(structuredContent = true)` method returns a parameterized type like `List<MyPojo>`, the output schema generation loses the generic type parameter. 

The generated output schema will look like:
```json
"outputSchema" : {
   "type" : "array",
   "items" : { }
}
```

A [custom output schema generator](https://docs.quarkiverse.io/quarkus-mcp-server/dev/guides-using-structured-content.html#_custom_schema_generator) can be used as a workaround.

mkouba:
- OutputSchemaGenerator now accepts java.lang.reflect.Type
- the Class variant is deprecated
- ToolDefinition.generateOutputSchema(Type) added for the programmatic API
- also fixed McpTestClientBase to handle JSON array structuredContent
- fixes #894


mkouba:
@angelozerr It would be great if you could test this fix with your setup.

PR #901, opened 2026-07-10
docs: DNS rebinding protection, HTTP troubleshooting, and connecting clients

rolfedh:
## What this does

Adds user-facing documentation for the Streamable HTTP transport's DNS rebinding protection and the connection problems it can produce. Today the only documentation for the feature is the generated configuration table and a release-notes bullet.

The motivation was a real support case: a user's MCP client could not connect, the server returned `403 Invalid origin` from the origin check, and the client reported it as "authentication required". With no prose docs, the cause was hard to find. A corporate HTTP proxy returning `403` produced the same client-side symptom.

## Changes

Four commits, documentation only:

1. **Security reference** - a "DNS Rebinding Protection" section: what it protects against, its scope (the Streamable HTTP endpoint), when it is active, what a rejection looks like, how to disable it, and how it relates to the CORS filter. The release-notes bullet now links to it.
2. **Transports concept** - a "Choosing a Transport" subsection contrasting the stdio and HTTP deployment models.
3. **Troubleshooting guide** (new page) - organized by symptom, with a table that separates the causes of a `403` (origin check, HTTP proxy, authentication or authorization) and curl commands to reproduce the origin check.
4. **Connecting MCP clients guide** (new page) - pointing Claude Code and other clients at an HTTP server, and the localhost-versus-remote implications.

## Verification

Every behavioral claim was checked against a running server (2.0.0.Beta3) and the source before it was written:

- Origin check behavior in `HttpMcpServerRecorder` (`403` with reason phrase `Invalid origin` and an empty body; requests without an `Origin` header pass; the `Non-localhost origin rejected` DEBUG log line and its category).
- Dev-versus-prod host default: `localhost` in dev and test mode (check active), `0.0.0.0` in production (check inactive).
- Auth-failure response shapes captured from a live server (a `401` challenge with `WWW-Authenticate`, and a `403` carrying a JSON-RPC error).
- `DnsRebindingTest` and `DnsRebindingDisabledTest` pass.

## Notes for reviewers

- No files under `docs/modules/ROOT/pages/includes/` were touched; those are generated by the config-doc plugin.
- The docs describe only the set of allowed localhost origins, not the matching implementation.
- The full build runs even though the diff is docs-only, because the `paths-ignore` globs in `build.yml` match root-level files only.
- Commit 4 (the client-connection guide) is the most opinion-prone part. Happy to drop or fold it if you would rather not carry client-facing content.

Follows up on a discussion with @mkouba about documenting this feature.


rolfedh:
> Commit 4 (the client-connection guide) is the most opinion-prone part. Happy to drop or fold it if you would rather not carry client-facing content.

@mkouba, the commits are split intentionally for review. Once you've decided on commit 4, I'll squash them into a single commit. You're also welcome to squash-merge if that's easier on your end.

rolfedh:
FYI @sberyozkin

docs/modules/ROOT/pages/reference-security.adoc
```diff
@@ -74,6 +74,51 @@ quarkus.http.cors.origins=http://localhost:6274 <2>
...
+
+[source,properties]
+----
+quarkus.mcp.server.http.dns-rebinding-check.enabled=false
+quarkus.mcp.server."my-server".http.dns-rebinding-check.enabled=false
+----
+
+[IMPORTANT]
+====
+This property is fixed at build time.
```
mkouba: Oh, this is actually a bug - the config phase for `McpHttpServersRuntimeConfig` should not be `BUILD_AND_RUN_TIME_FIXED` but obviously `RUN_TIME`.

I've created https://github.com/quarkiverse/quarkus-mcp-server/issues/904.

We need to fix that first.

mkouba: Ok, fixed. @rolfedh Could you pls rebase and update the PR accordingly?

mkouba: @rolfedh ping

rolfedh:
@mkouba Rebased onto `main` and updated the PR for the #904 fix.

The security reference no longer says that `quarkus.mcp.server.http.dns-rebinding-check.enabled` is fixed at build time. The admonition is replaced with: "The property is read at startup, so you can override it at run time, for example with an environment variable or a system property." I verified this against `HttpMcpServerRecorder`, which reads the value once when it creates the endpoint handler.

That admonition was the only place in the PR that mentioned build-time behavior, so no other pages needed changes. Ready for another look.

rolfedh:
Squashing to a single commit now.

mkouba (approved):
LGTM

PR #905, opened 2026-07-13
http: fix incorrect config phase for McpHttpServersRuntimeConfig

### LINKED ISSUE #904: McpHttpServersRuntimeConfig has incorrect config phase
It should be `RUNTIME` instead of `BUILD_AND_RUN_TIME_FIXED`.

Note that this should not be a breaking change. The fix only unlocks the ability to change the relevant properties at runtime.

mkouba:
- McpHttpServersRuntimeConfig used BUILD_AND_RUN_TIME_FIXED but all its properties (dns-rebinding-check.enabled, auto-init, lazy-sse-init) are purely runtime in nature
- changed to RUN_TIME to match the other runtime configs and allow these properties to be overridden at startup.
- resolves #904

### DIFF
--- a/docs/modules/ROOT/pages/includes/quarkus-mcp-server-http.adoc
+++ b/docs/modules/ROOT/pages/includes/quarkus-mcp-server-http.adoc
@@ -88,7 +88,7 @@ endif::add-copy-button-to-env-var[]
 |boolean
 |`+++false+++`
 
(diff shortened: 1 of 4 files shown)


mkouba:
Note that this should not be a breaking change. The fix only unlocks the ability to change the relevant properties at runtime.

PR #926, opened 2026-08-03
Properly cancel subscription on notifications/canceled over stdio

### LINKED ISSUE #925: notifications/cancelled does not cancel subscriptions/listen subscriptions in stateless stdio transport
When using the stateless protocol (2026-07-28) over stdio, sending notifications/cancelled referencing a subscriptions/listen request ID fails to cancel the subscription. Notifications continue to be delivered after the cancellation.

The cause is in `McpMessageHandler.cancelRequest()` it calls `mcpRequest.connection().removeSubscription(requestId)`, but in stateless mode each incoming message is processed on a new transient connection. The subscription was added to the transient connection created for the original `subscriptions/listen` message, while the `notifications/cancelled` arrives on a different transient connection that has no subscriptions. `removeSubscription()` returns false and the cancellation is silently ignored. Per the spec, `notifications/cancelled` is the cancellation mechanism for stdio (HTTP uses SSE stream closure instead). The fix needs to search across all connections in the ConnectionManager when the subscription is not found on the current request's connection.

jmartisk:
Fixes https://github.com/quarkiverse/quarkus-mcp-server/issues/925

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/McpMessageHandler.java
+++ b/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/McpMessageHandler.java
@@ -475,7 +475,7 @@ public abstract class McpMessageHandler<MCP_REQUEST extends McpRequest> {
                 String reason = params.getString("reason");
                 LOG.debugf("Cancel request with id %s: %s [%s]", requestId, reason != null ? reason : "no reason",
                         mcpRequest.connection().id());
-            } else if (requestId != null && mcpRequest.connection().removeSubscription(requestId)) {
+            } else if (requestId != null && cancelSubscription(requestId, mcpRequest)) {
                 LOG.debugf("Subscription %s cancelled [%s]", requestId, mcpRequest.connection().id());
             } else {
                 LOG.warnf("Ignored unknown/completed/invalid cancel request with id %s [%s]", requestId,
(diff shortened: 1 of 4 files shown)


core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/McpMessageHandler.java
```diff
@@ -485,6 +485,24 @@ private Future<Void> cancelRequest(JsonObject message, MCP_REQUEST mcpRequest) {
         return Future.succeededFuture();
     }
 
+    private boolean cancelSubscription(Object requestId, MCP_REQUEST mcpRequest) {
+        // First try the current request's connection
+        if (mcpRequest.connection().removeSubscription(requestId)) {
+            return true;
+        }
```
mkouba:
```suggestion
        } else if (transport() == Transport.STDIO) {
         // For stateless protocols...
```
Because the iteration makes no sense for other types of transports?

mkouba (commented):
Looks good! I've added one minor comment.

Thanks!

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/McpMessageHandler.java
jmartisk: applied

mkouba (approved):

mkouba:
Thanks!

PR #945, opened 2026-08-11
core: add targeted notifyListChanged(Predicate<McpConnection>) to feature managers

### LINKED ISSUE #943: Targeted (single-connection) notifications/tools/list_changed
## Use case / problem

We're building an MCP server using `quarkus-mcp-server-http`, exposing a fairly large and growing tool surface across many categories. We already use `ToolFilter` (per #248/#257) to scope each connection's `tools/list` to only the categories a given user/session is authorized for, which keeps typical sessions well under the full tool count.

We'd like to go a step further: even within what a session is *authorized* for, default to exposing only a small, commonly-used subset, plus two always-visible meta-tools:
- one to list the currently-disabled tools in a given category, so the agent can discover what exists without it being loaded into context up front, and
- an `enable_tool` tool the agent calls to actually activate a specific disabled tool — either for a single subsequent call (auto-disabled again immediately after use) or for the rest of the current session.

The goal is to keep the default tool list small without limiting what's actually reachable.

The blocker: `notifications/tools/list_changed` is only ever sent via `FeatureManagerBase#notifyConnections`, which broadcasts to **every** connection (`ToolManagerImpl#newTool`/`removeTool` are the only call sites). Our activation is per-connection (driven by `ToolFilter` state, not by structurally registering/removing tools), so we have no way to tell just the one session whose filter output changed to re-fetch its list — the only lever we have would notify every other connected session too.

We looked at #622 and found the actual precedent for what we're asking: `ResourceManagerImpl#sendUpdateNotifications` already looks up one specific connection by ID (`connectionManager.get(connectionId)`) and calls `connection.send(updated)` directly on it. That capability just isn't exposed publicly — `McpConnection`'s public interface has no `send()`, and `ConnectionManager` lives in the internal `io.quarkiverse.mcp.server.runtime` package.

## Proposed API

Something like:

```java
public interface ToolManager extends FeatureManager<ToolInfo> {
    ...
    /**
     * Sends a {@code notifications/tools/list_changed} notification to the given
     * connection only, without registering/removing any tool.
     */
    void notifyListChanged(McpConnection connection);
}
```

`ToolFilter#test(ToolInfo, McpConnection)` (or the `FilterContext` variant) already receives the connection, so the reference needed to call this is already in hand wherever the filter's backing state is mutated — e.g. inside the `enable_tool` tool's implementation.

## Scope

**In scope for this request:**
- A single-connection-targeted send of `notifications/tools/list_changed`, following the same pattern already used internally by `ResourceManagerImpl#sendUpdateNotifications`.

**Explicitly out of scope / not needed from us:**
- Any change to `ToolFilter` itself — the filtering half of this already works well for our case.
- Elicitation — unrelated existing feature, not part of this request.
- The single-use "enable once, then auto-disable" flow — pure reuse of the same targeted-notify primitive twice (enable, then disable after the call), no separate API needed.

**Worth naming now as possible future additions, not requested here:**
- Fan-out to *all* of one identity's currently-open connections, for a user with multiple concurrent sessions who wants a tool enabled "for all my sessions," not just the one that asked. Right now `ConnectionManager` is internal, so there's no public way to even enumerate a user's other live connections — this would need its own, larger API and isn't something we currently need.
- A connection-lifecycle hook (e.g. on close/disconnect), so an application can evict any session-scoped state (like our per-connection "enabled tools" set) when a connection ends. We have no such signal via the public API today. Not blocking for the core request here, but related, and would probably be its own issue if pursued.

## Offer

We'd be happy to implement this ourselves and open a PR if the shape above looks reasonable to you — just let us know if you'd rather take a different approach to the API before we start.

mkouba:
- adds a way to send list_changed notifications to a subset of connections, selected by a predicate
- this is useful when the effective feature list changes for specific connections, e.g. due to a filter state change
- resolves #943

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/mcp/server/PromptManager.java
+++ b/core/runtime/src/main/java/io/quarkiverse/mcp/server/PromptManager.java
@@ -2,6 +2,7 @@ package io.quarkiverse.mcp.server;
 
 import java.util.List;
 import java.util.Map;
+import java.util.function.Predicate;
 
 import io.quarkiverse.mcp.server.PromptManager.PromptInfo;
 
@@ -59,6 +60,17 @@ public interface PromptManager extends FeatureManager<PromptInfo> {
      * @see #removePrompt(String, String)
      */
     PromptInfo removePrompt(String name);
+
+    /**
+     * Sends a {@code notifications/prompts/list_changed} notification to the connections that match the given filter.
+     * <p>
+     * Unlike {@link PromptDefinition#register()} and {@link #removePrompt(String)}, which broadcast to all connections,
+     * this method allows targeted notification delivery. This is useful when the effective prompt list changes for specific
+     * connections only, for example due to a {@link PromptFilter} state change.
+     *
+     * @param filter the predicate used to select connections to notify
+     */
+    void notifyListChanged(Predicate<McpConnection> filter);
 
     /**
      * Prompt info.
--- a/core/runtime/src/main/java/io/quarkiverse/mcp/server/ResourceManager.java
+++ b/core/runtime/src/main/java/io/quarkiverse/mcp/server/ResourceManager.java
@@ -3,6 +3,7 @@ package io.quarkiverse.mcp.server;
 import java.util.Map;
 import java.util.Optional;
 import java.util.OptionalInt;
+import java.util.function.Predicate;
 
 import io.quarkiverse.mcp.server.ResourceManager.ResourceInfo;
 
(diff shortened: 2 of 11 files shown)


b-eyselein (approved):
Reviewed and tested — LGTM.

Review: the implementation matches what was discussed, is minimal, and reuses the existing broadcast machinery cleanly (`notifyConnections(method, filter)` as an overload of the existing `notifyConnections(method)`). No correctness concerns.

Testing: checked out the PR branch and ran the full `core-deployment` and `transports/http/deployment` test suites (359 tests) — everything passes, no regressions.

For our own use case, this covers stateful precisely, and stateless well enough for now via a broadcast-style filter (`conn -> true`) — we don't control the clients connecting to our server, so we can't rely on any client-side cooperation to correlate a specific stateless connection back to the one that triggered the change. We do have an authenticated identity available server-side that could enable precise per-user targeting if `McpConnection` exposed something to match on, but we'll treat that as a separate future issue if it turns out we actually need it rather than asking for it now.

Thanks for turning this around so quickly.

PR #955, opened 2026-08-17
fix: Add default message to OperationCancellationException

kornys:
OperationCancellationException had no constructor, causing getMessage() to return null.
When wrapped in ToolCallException and converted to a ToolResponse via ToolResponse.error(tce.getMessage()), this resulted in IllegalArgumentException from TextContent which requires non-null text.

* Added no-arg constructor with "Operation cancelled" default message.
* Added test to verify cancellation doesn't throw IllegalArgumentException due to null message.

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/mcp/server/Cancellation.java
+++ b/core/runtime/src/main/java/io/quarkiverse/mcp/server/Cancellation.java
@@ -68,6 +68,10 @@ public interface Cancellation {
 
         private static final long serialVersionUID = 1L;
 
+        public OperationCancellationException() {
+            super("Operation cancelled");
+        }
+
     }
 
 }
--- a/transports/http/deployment/src/test/java/io/quarkiverse/mcp/server/test/cancel/CancellationTest.java
+++ b/transports/http/deployment/src/test/java/io/quarkiverse/mcp/server/test/cancel/CancellationTest.java
@@ -57,6 +57,30 @@ public class CancellationTest extends McpServerTest {
                 .thenAssertResults();
     }
 
+    @Test
+    public void testCancellationErrorMessage() throws InterruptedException {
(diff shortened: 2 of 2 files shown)


mkouba:
Thanks for the pull request. I have a question - how and why do you wrap the `OperationCancellationException` in a `ToolCallException`? You're not using the `@WrapBusinessError` because there the exception message should be something like `io.quarkiverse.mcp.server.Cancellation$OperationCancellationException`.

Also the added test doesn't seem to reproduce the bug - with no `ToolCallException` wrapping, it goes straight through the `MessageHandler`, no `TextContent` is created. Actually, the `OperationCancellationException` is not a business error, so it should not result in a tool error.

Maybe we should just fix the `WrapBusinessErrorInterceptor.wrapIfNecessary()` to never wrap the cancellation exception...

kornys:
@mkouba yeah I wrongly identified the issue. I use `@WrapBusinessError(value = Exception.class, unless = ToolCallException.class)`. I can fix it in strimzi-mcp by adding also `unless = {ToolCallException.class, Cancellation.OperationCancellationException.class}` or if I can as you suggest implement part to never wrap cancellation exception.

mkouba:
I think that `WrapBusinessErrorInterceptor` should never wrap `ToolCallException`, `OperationCancellationException`, `org.mcpjava.server.Cancellation.OperationCancelledException`, `InputRequiredException` and `UrlElicitationRequiredException`. These all are control-flow signals that are handled specially.

On the other hand, it's a breaking change and if we hard-exclude those, there will be no way to opt-out... which shouldn't be a problem because there's no reason to wrap these anyway (at least I can't see any).

Would you care to implement it?

kornys:
Yes I will do that

kornys:
closing this PR as suggested implementation is opened in #959

PR #959, opened 2026-08-18
refactor: never wrap control-flow exceptions in WrapBusinessErrorInterceptor

kornys:
Prevents `WrapBusinessErrorInterceptor` from wrapping control-flow exceptions that are handled specially by `MessageHandler`, even when they match `@WrapBusinessError` annotation filters.

This resolves the issue identified in #955 where using `@WrapBusinessError(value = Exception.class, unless = ToolCallException.class)` would wrap `OperationCancellationException`.

## Control-Flow Exceptions Never Wrapped

  1. `ToolCallException` - Already a tool error wrapper (prevents double-wrapping)
  2. `Cancellation.OperationCancellationException` - Cancellation signal
  3. `org.mcpjava.server.Cancellation.OperationCancelledException` - mcpjava cancellation variant
  4. `InputRequiredException` - Multi-round-trip input request
  5. `UrlElicitationRequiredException` - URL elicitation request

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/mcp/server/InputRequiredException.java
+++ b/core/runtime/src/main/java/io/quarkiverse/mcp/server/InputRequiredException.java
@@ -48,9 +48,10 @@ public class InputRequiredException extends RuntimeException {
     }
 
     /**
+     * @return a new builder
      * @see MrtrRequest#inputRequired()
      */
-    static Builder builder() {
+    public static Builder builder() {
         return new Builder();
     }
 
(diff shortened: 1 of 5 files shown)


mkouba (commented):
It looks good but I'm missing test coverage for `InputRequiredException`, `UrlElicitationRequiredException` and `ToolCallException`.

In fact, we could add a plain unit test for `WrapBusinessErrorInterceptor` to test the behavior of `wrapIfNecessary()`. WDYT?

mkouba:
Also squash the commits pls :pray:.

kornys:
> It looks good but I'm missing test coverage for `InputRequiredException`, `UrlElicitationRequiredException` and `ToolCallException`.
> 
> In fact, we could add a plain unit test for `WrapBusinessErrorInterceptor` to test the behavior of `wrapIfNecessary()`. WDYT?

yeah unit tests for this makes sense, I ll do that and squash that as well

kornys:
I have added unit tests. I didn't want to change method wrapIfNecessary prom private to package private so I had to do reflections

mkouba:
> I have added unit tests. I didn't want to change method wrapIfNecessary prom private to package private so I had to do reflections

It is actually a good idea to make it package private (and static), and get rid of all reflection from the test. I will try to push a commit to your PR.

PR #979, opened 2026-08-27
Use canonical `ai` instead of `artificial-intelligence` as category

### LINKED ISSUE #978: MCP category should be "ai" not "artificial-intelligence"
The correct category for AI things is `ai`. 

"But wait, "you say, "what's on code.quarkus.io is `artificial-intelligence`". Yeah, that's https://github.com/quarkusio/code.quarkus.io/issues/1836. 

"But how am I supposed to work out the right category?", you say. Yeah, you can't, not really. https://github.com/quarkusio/quarkus/issues/55974 covers all that.

You can see Langchain4J is using `ai` – https://quarkus.io/extensions/?categories=ai. It's also what's in the magic categories file in the quarkus repo(but it's not exactly what's in the registry, if you're being picky, because https://github.com/quarkusio/quarkus/issues/55974).

Fix incoming.

holly-cummins:
Resolves #978

### DIFF
--- a/core/runtime/src/main/resources/META-INF/quarkus-extension.yaml
+++ b/core/runtime/src/main/resources/META-INF/quarkus-extension.yaml
@@ -5,6 +5,6 @@ metadata:
     - mcp
   guide: https://docs.quarkiverse.io/quarkus-mcp-server/dev/
   categories:
-    - "artificial-intelligence"
+    - "aigence"
   status: "stable"
   unlisted: true
--- a/core/runtime/src/main/resources/META-INF/quarkus-extension.yaml
+++ b/core/runtime/src/main/resources/META-INF/quarkus-extension.yaml
@@ -5,6 +5,6 @@ metadata:
     - mcp
   guide: https://docs.quarkiverse.io/quarkus-mcp-server/dev/
   categories:
-    - "aigence"
+    - "ai"
   status: "stable"
   unlisted: true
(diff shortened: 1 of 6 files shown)


core/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -5,6 +5,6 @@ metadata:
     - mcp
   guide: https://docs.quarkiverse.io/quarkus-mcp-server/dev/
   categories:
-    - "artificial-intelligence"
+    - "aigence"
```
mkouba: @holly-cummins The title of the PR says `ai` not `aigence` :D

holly-cummins: OMG. My IDE seems to have gone a big rogue. "How hard can a simple search and replace be?" Very hard, apparently! 

Will fix.

holly-cummins: I fixed, but left the original for comedy value. "Comedy value" maybe isn't a design goal, so let me know if you'd prefer I squashed. :)

mkouba: The "comedy value" is indeed not a goal, but I like it. Let's keep it this way. ;-)

mkouba (approved):

holly-cummins:
@mkouba, could this be backported to the 2.x release? 

Getting it into a release should:
- Reduce a bit of clutter on quarkus.io/extensions
- Get the MCP server turning up in code.quarkus.io when people [filter by the AI category](https://code.quarkus.io/?extension-search=category:artificial-intelligence)

PR #984, opened 2026-09-01
core: scope list-changed notifications to the affected servers

### LINKED ISSUE #983: ToolManager.register()/removeTool() always broadcast tools/list_changed to every connection, including connections of other servers
### Version

`quarkus-mcp-server` 2.0.0, `quarkus-mcp-server-http`, Quarkus 3.36.0, JDK 25.

### Summary

`ToolManager.newTool(...).register()` and `ToolManager.removeTool(...)` unconditionally send `notifications/tools/list_changed` to **every open connection in the process**, with no filtering by server name. In a deployment that serves multiple MCP servers from one application, registering a tool on server A also notifies every client connected to servers B, C, … , even though their tool list did not change.

This is especially surprising for `removeTool(String name, String serverName)`, which is explicitly scoped to a single server and still notifies globally.

### Why this matters

We run a multi-tenant integration platform where each tenant is a separate MCP server name, and tools are registered/removed at runtime as customers publish and unpublish flows. Two consequences:

1. **Wasted round trips.** Every connected tenant re-fetches `tools/list` whenever *any* other tenant publishes a flow. With many tenants and frequent publishing, this amplifies badly.
2. **Cross-tenant signal leak.** A tenant can infer that *some other* tenant registered or removed a tool, purely from the timing of the notifications it receives. No content leaks, but the activity signal does — and in a multi-tenant product that is hard to justify to a security review.

### Relationship to #943

[#943](https://github.com/quarkiverse/quarkus-mcp-server/issues/943) added `ToolManager.notifyListChanged(Predicate<McpConnection>)`, which is great and solves the case where the *effective* list changes without registration (e.g. a `ToolFilter` state change). We use it and it works correctly.

But it does not cover this case: the broadcast is emitted from **inside** `register()` / `removeTool()`, and there is no way to suppress or scope it. So a multi-tenant application cannot avoid the global notification on the registration path, no matter how carefully it uses `notifyListChanged`.

### Where it happens

`ToolManagerImpl.register()`:

```java
notifyConnections(McpMethod.NOTIFICATIONS_TOOLS_LIST_CHANGED);
toolAddedEvent.fire(new ToolAdded(ret));
return ret;
```

`ToolManagerImpl.removeTool(String name, String serverName)` — note `serverName` is known here:

```java
tools.computeIfPresent(key, (k, value) -> {
    if (!value.isMethod()) {
        removed.set(value);
        notifyConnections(McpMethod.NOTIFICATIONS_TOOLS_LIST_CHANGED);
        return null;
    }
    return value;
});
```

Both reach the unfiltered overload in `FeatureManagerBase`:

```java
protected void notifyConnections(McpMethod method) {
    notifyConnections(method, null);
}

protected void notifyConnections(McpMethod method, Predicate<McpConnection> filter) {
    ...
    for (McpConnectionBase c : connectionManager) {
        if (c.status() != McpConnection.Status.IN_OPERATION) continue;
        if (filter != null && !filter.test(c)) continue;
        ...
    }
}
```

The loop iterates the whole `ConnectionManager` and never compares `c.serverName()` against the server(s) the tool was registered for.

### Reproduction

Configure two MCP servers, then:

1. Open a `subscriptions/listen` stream on **server A** with `{"notifications":{"toolsListChanged":true}}`.
2. Open the same stream on **server B**.
3. At runtime, register a tool bound only to server A:
   `toolManager.newTool("x").setServerName("A").setDescription("…").setHandler(…).register()`

**Expected:** only the stream on A receives `notifications/tools/list_changed`.
**Actual:** both streams receive it.

Observed on 2.0.0. For contrast, in the same setup
`notifyListChanged(conn -> "A".equals(conn.serverName()))` behaves correctly — A receives it, B does not.

<details>
<summary>Note on our test setup</summary>

Our own measurement used a wildcard Vert.x route that dispatches `/mcp/:serverId` to `StreamableHttpMcpMessageHandler`, because our server names only exist at runtime. That detail should not matter here: `notifyConnections` iterates `ConnectionManager` with no server-awareness at all, so a plain multi-server configuration hits the same path. Happy to put together a self-contained reproducer against the standard configuration if that would help.

</details>

### Possible solutions

Any of these would work for us; listed roughly by increasing invasiveness:

1. **Scope the built-in broadcast to the affected servers.** Notify only connections whose `serverName()` is in the registered/removed tool's `serverNames()`. Arguably the correct behavior, but it is a behavior change for anyone relying on the current broadcast.
2. **Opt out at registration.** e.g. `ToolDefinition#register(boolean notify)` / a `setNotifyListChanged(false)` on the builder, plus an equivalent `removeTool` overload — leaving the caller to invoke `notifyListChanged(...)` with the right filter.
3. **Opt into a filter.** e.g. `register(Predicate<McpConnection> notifyFilter)`, mirroring the shape already introduced by #943.

We would be glad to open a PR for whichever direction you prefer — happy to be told which one fits the project's compatibility expectations before writing any code.

### Current workaround

None that removes the noise. We use `notifyListChanged` with a server-name filter where we control the trigger, and accept the extra cross-tenant notifications on the register/remove path.

mkouba:
- registering or removing a tool, prompt, resource or resource template programmatically broadcast notifications/*_list_changed to every open connection in the process, regardless of the server the connection was bound to
- in multi-server deployments this caused wasted round trips and leaked cross-server activity signals
- the automatic notification is now scoped to the connections of the servers the affected feature is bound to
- a new config property quarkus.mcp.server.auto-list-changed-strategy can be used to restore the previous behavior
- fixes #983

### DIFF
--- a/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/FeatureManagerBase.java
+++ b/core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/FeatureManagerBase.java
@@ -61,6 +61,7 @@ import io.quarkiverse.mcp.server.Sampling;
 import io.quarkiverse.mcp.server.TransportHint;
 import io.quarkiverse.mcp.server.runtime.ResultMappers.Result;
 import io.quarkiverse.mcp.server.runtime.config.McpServersRuntimeConfig;
+import io.quarkiverse.mcp.server.runtime.config.McpServersRuntimeConfig.AutoListChangedStrategy;
 import io.quarkiverse.mcp.server.runtime.config.McpServersRuntimeConfig.InvalidServerNameStrategy;
 import io.quarkiverse.mcp.server.runtime.mcpjava.McpJavaCancellationAdapter;
 import io.quarkiverse.mcp.server.runtime.mcpjava.McpJavaMcpRequestAdapter;
@@ -91,6 +92,9 @@ public abstract class FeatureManagerBase<RESULT, INFO extends FeatureManager.Fea
     // used to validate server name if InvalidServerNameStrategy.FAIL is set
     protected final Set<String> serverNames;
 
+    // determines which connections receive the automatic list_changed notification on register/remove
+    protected final AutoListChangedStrategy autoListChangedStrategy;
+
     // used to ensure atomic registration of features across multiple server keys
     protected final Lock registrationLock = new ReentrantLock();
 
@@ -111,6 +115,7 @@ public abstract class FeatureManagerBase<RESULT, INFO extends FeatureManager.Fea
         this.serverRequests = serverRequests;
         this.cancellationRequests = cancellationRequests;
         this.serverNames = config.invalidServerNameStrategy() == InvalidServerNameStrategy.FAIL ? metadata.serverNames() : null;
+        this.autoListChangedStrategy = config.autoListChangedStrategy();
         jakarta.enterprise.inject.Instance<McpTracing> mcpTracingInstance = Arc.container().select(McpTracing.class);
         this.mcpTracing = mcpTracingInstance.isResolvable() ? mcpTracingInstance.get() : null;
     }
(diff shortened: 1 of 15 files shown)


mkouba:
CC @alexandredavi

PR #1005, opened 2026-09-11
core: introduce McpResponseException hierarchy

### LINKED ISSUE #1004: Introduce McpResponseException
The `io.quarkiverse.mcp.server.McpException` is currently automatically converted to a JSON-RPC error message.

We could introduce `io.quarkiverse.mcp.server.McpResponseException` - an abstract exception whose subtypes would be automatically converted to a JSON-RPC response.

mkouba:
- add McpResponseException as the abstract supertype for exceptions that are automatically converted to a JSON-RPC response
- McpException renders as an error (unchanged)
- McpResultException supplies a success result
- McpException and InputRequiredException now extend these types
- WrapBusinessErrorInterceptor uses a single McpResponseException guard
- resolves #1004


mkouba:
Hm, the `McpResultException.result()` returns `Object` and explicitly permits maps, JSON objects, and arbitrary JSON-serializable objects. However, `Sender.sendResult()` modifies supplied maps and JSON objects to add `resultType` and merge response metadata :facepalm:.

PR #1008, opened 2026-09-16
Use canonical ai instead of artificial-intelligence as category

mkouba:
Resolves #978


mkouba:
This will be included in 2.0.2.

PR #1012, opened 2026-09-17
core: initial support for MCP extensions

### LINKED ISSUE #1000: Support for MCP extension capabilities and custom methods (needed for SEP-2640 "Skills")
MCP recently finalized SEP-2640, the "Skills" extension (https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/seps/2640-skills-extension.mdx). 

Like other MCP extensions, it's designed to be additive on top of the base protocol — a server declares support for it and implements a couple of extra pieces:

A new capability advertised in the initialize response

```json
"capabilities": {
   "extensions": {
     "io.modelcontextprotocol/skills": { "directoryRead": true }
   }
 }
```

2. Two new top-level JSON-RPC methods, skills/list and skills/get (and optionally resources/directory/read).

The MCP spec is clearly designed to let extensions like this be adopted incrementally by servers/SDKs ahead of (or independent from) core spec support, so we'd expect this to be something an application built on quarkus-mcp-server could add on its own.

As far as we can tell, quarkus-mcp-server currently has no extension point for either of these:

 - There's no way for an application to add its own top-level JSON-RPC method (like skills/list) alongside the built-in ones (tools/*, resources/*, etc.) — unrecognized methods are simply rejected.
 - There's no way for an application to contribute an additional key (like extensions) to the capabilities object returned in initialize — that object seems to be fully owned by the extension with a fixed set of keys.

Would you consider adding:

- A way for applications to register additional custom JSON-RPC methods handled by their own code
- A way for applications to contribute extra keys into the server's advertised capabilities (e.g. an extensions map)

SEP-2640 "Skills" over MCP is my main use case

mkouba:
- implements infrastructure for MCP extensions
- McpExtension marks a bean as an extension; its id (a _meta-style key with a mandatory prefix) and the settings object declared via McpExtensionSetting are advertised under capabilities.extensions in the initialize/server/discover responses
- McpExtensionMethod registers custom top-level JSON-RPC methods handled like features (full provider + named-param binding, sourced from the top-level params object; sync or async, POJO/JsonObject/Map results)
- McpServer bindings are honored per extension (advertising + dispatch); bindings are class-level only - McpServer on a method is rejected
- build-time validation: invalid/prefixless id, duplicate id, duplicate method, built-in method clash, method outside an extension
- resolves #1000


mkouba:
CC @jonesbusy @matheusandre1

jonesbusy:
Thanks! 

Tested a SNAPSHOT version of the first commit against a real Quarkus MCP server, and it works well (Didn't check the code, just the usage) — I was able to fully implement SEP-2640 "Skills" (both the capability advertisement and the skills/list/skills/get custom methods) without touching the extension itself. Confirmed working end-to-end with MCP Inspector's new Skills tab.

For example

```java
@ApplicationScoped
@McpExtension(id = "io.modelcontextprotocol/skills")
@McpExtensionSetting(name = "directoryRead", type = MetaField.Type.BOOLEAN, value = "false")
public class SkillsExtension {

    ....

    @McpExtensionMethod("skills/list")
    public SkillsListResult list() {
        List<SkillEntry> skills = registry.allSkills().stream()
                .map(this::toEntry)
                .toList();
        return new SkillsListResult("complete", skills, 60_000L, "private");
    }

    @McpExtensionMethod("skills/get")
    public SkillGetResult get(String uri) {
        SkillEntry skill = registry.allSkills().stream()
                .map(this::toEntry)
                .filter(entry -> entry.uri().equals(uri))
                .findFirst()
                .orElseThrow(() -> new McpException("Unknown skill: " + uri, JsonRpcErrorCodes.INVALID_PARAMS));
        return new SkillGetResult("complete", skill);
    }

....

}
```

The @McpExtension/@McpExtensionMethod/@McpExtensionSetting annotations did exactly what I needed — capability negotiation, custom top-level JSON-RPC dispatch, provider injection (McpConnection etc.) all worked as documented. The build-time validation (d also kicked in correctly.

MCP insepector is also happy about it

Skills_Over_MCP

mkouba:
> Thanks!
> 
> Tested a SNAPSHOT version of the first commit against a real Quarkus MCP server, and it works well (Didn't check the code, just the usage) — I was able to fully implement SEP-2640 "Skills" (both the capability advertisement and the skills/list/skills/get custom methods) without touching the extension itself. Confirmed working end-to-end with MCP Inspector's new Skills tab.

@jonesbusy That's great! I did some minor tweaks and added docs as well. So the PR should be ready now.

Do you think it would make sense to introduce the `quarkus-mcp-server-skills` module?

jonesbusy:
Perhaps. If it simply usage for consumers.

My only concern is where the MCP is suppose to discover skills to expose? On the filesystem?

On my side my skills are fetched from an OCI internal registry from the MCP (using user identity) and exposed.

If `quarkus-mcp-server-skills` ever exists it also need a clear API to provide skills data to it.

mkouba:
> Perhaps. If it simply usage for consumers.
> 
> My only concern is where the MCP is suppose to discover skills to expose? On the filesystem?

Hm, this would need to be pluggable. Maybe an interface that could be implemented by a CDI bean so that it's easy to consume. And maybe some configurable built-in impls (filesystem)?
> 
> On my side my skills are fetched from an OCI internal registry from the MCP (using user identity) and exposed.
> 
> If `quarkus-mcp-server-skills` ever exists it also need a clear API to provide skills data to it.

Exactly.

PR #1013, opened 2026-09-18
Introduce MCP Tasks support

### LINKED ISSUE #900: MCP Tasks support
It would be really great if Quarkus MCP Server could support [Tasks](https://modelcontextprotocol.io/extensions/tasks/overview)

In my MCP Server which install, starts LSP/DAP server, initialize (indexing step) LSP/DAP server, and for LSP execute an LSP feature (ex: references), I could create a task per those steps and it will give a nice overview for the user.

For now I am trying to do that with Progress.

matheusandre1:
Closes: #900 

- Adds built-in support for the [MCP Tasks](https://modelcontextprotocol.io/extensions/tasks/overview) extension (`io.modelcontextprotocol/tasks`), built on top of the extensions infrastructure from #1000.

### API

- `@Task` marks a `@Tool` method as task-augmented (`required`, `ttl`, `pollInterval`).
- `TaskContext` parameter: status message updates and `inputRequestBuilder()` to request
  elicitation/sampling/roots input while the task is running.
- `TaskInputRequest`, `TaskStatus`, `TaskManager` (inspect/cancel tasks).
- `ToolManager.TaskOptions` / `ToolDefinition#setTaskOptions()` for programmatic tools.
- Config: `quarkus.mcp.server.tasks.default-ttl` (1h), `quarkus.mcp.server.tasks.default-poll-interval` (5s).


mkouba:
I think that the MCP Tasks extension should go in a separate module, i.e. `/extensions/tasks`.

core/runtime/src/main/java/io/quarkiverse/mcp/server/McpServer.java
```diff
@@ -40,6 +40,12 @@
      */
     String DEFAULT = "<default>";
 
+    /**
+     * A special value that binds the annotated element to <em>all</em> server configurations known at build time. It is
+     * intended for features and extensions provided by libraries.
+     */
+    String ALL = "*";
```
mkouba, line 47: I think that we should extract this functionality in a separate PR with dedicated test coverage and updated docs. It's actually an interesting feature.

I'm not so sure about the constant value though - something like `<all>` could make more sense.

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/ToolCallInterceptor.java
```diff
@@ -0,0 +1,86 @@
...
+import io.vertx.core.json.JsonObject;
+
+/**
+ * Intercepts a {@code tools/call} request before the tool is executed.
+ * <p>
+ * This is an internal extension point used by MCP extensions that need to alter the standard handling of a tool call, e.g.
+ * to respond with a different result shape and execute the tool asynchronously (MCP Tasks). Implementations are CDI beans.
+ * Interceptors are only consulted for tools that exist for the server of the request and pass the tool filters.
+ */
+public interface ToolCallInterceptor {
```
mkouba, line 17: I think that we should extract this internal API into a separate PR as well. And it should be probably part of the public API because other MCP extensions might require this functionality as well.

I'll try to think about it a bit.

core/runtime/src/main/java/io/quarkiverse/mcp/server/McpServer.java
mkouba: See https://github.com/quarkiverse/quarkus-mcp-server/issues/1018

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/ToolCallInterceptor.java
mkouba: I think that first we need a way to contribute to the list of injected parameters of a feature method - that would be the idiomatic API for the moment (`Elicitation`, etc.). See https://github.com/quarkiverse/quarkus-mcp-server/issues/1019.

core/runtime/src/main/java/io/quarkiverse/mcp/server/McpServer.java
mkouba: I've just merged the PR that resolves https://github.com/quarkiverse/quarkus-mcp-server/issues/1018.

core/runtime/src/main/java/io/quarkiverse/mcp/server/runtime/ToolCallInterceptor.java
mkouba: Here's a DRAFT PR: https://github.com/quarkiverse/quarkus-mcp-server/pull/1021

matheusandre1: Sorry for the delay; I agree with the points...

- Once #1021 lands, I'll rework the module on top of it: `TaskContext` becomes a
  parameter contributed via `CustomArgumentBuildItem`/`McpArgumentProvider`, and I'll
  remove my `CustomArgumentTypeBuildItem`, `ToolCallInterceptor` and the
  `SubscriptionFilterExtension` core changes so that the core diff is reduced to
  what #1021 provides.
- For the API shape, I like the idea from #1019: a `Tasks` parameter with
  `isSupported()` and `create(...)` that sends the `CreateTaskResult` through the
  existing `McpResultException` seam. Two questions before I go that way:
  1. Should the asynchronous execution still be triggered by an annotation
     (`@Task`) for the common case, with `Tasks.create()` as the programmatic
     variant, or do you prefer `Tasks.create()` only?
  2. The task needs its own request context (the original request is over once
     the `CreateTaskResult` is sent) and a way to observe `tasks/cancel` via
     `Cancellation`. Is it OK if the tasks module manages a new duplicated
     Vert.x context for the task itself, or would you rather have core expose a
     helper for that?

mkouba: > Sorry for the delay

No problem at all.

> 1. Should the asynchronous execution still be triggered by an annotation
>        (`@Task`) for the common case, with `Tasks.create()` as the programmatic
>        variant, or do you prefer `Tasks.create()` only?

That's a good question. I would start with the programmatic API. Mainly because it's a pattern that we already use in the MCP server. I like the idea of `@Task` but we need to be careful here, because we'll have to reflect the [execution model](https://docs.quarkiverse.io/quarkus-mcp-server/dev/concepts-architecture.html#_execution_model) of a feature method. For `Tasks.create()` we can reserve a dedicated thread pool and specify rules for the submitted logic.

> 2\. The task needs its own request context (the original request is over once
>         the `CreateTaskResult` is sent) and a way to observe `tasks/cancel` via
>         `Cancellation`. Is it OK if the tasks module manages a new duplicated
>         Vert.x context for the task itself, or would you rather have core expose a
>         helper for that?

I think that we could reuse `io.quarkus.vertx.VertxContextSupport.executeBlocking()` helper in the tasks module.

mkouba: BTW I've renamed `CustomArgumentBuildItem`/`McpArgumentProvider` to `FeatureArgumentProviderBuildItem/FeatureArgumentProvider`, as it aligns with `FeatureManager`, `FeatureArguments`, `FeatureInfo`, etc.

mkouba: > > 2. The task needs its own request context (the original request is over once
> > the `CreateTaskResult` is sent) and a way to observe `tasks/cancel` via
> > `Cancellation`. Is it OK if the tasks module manages a new duplicated
> > Vert.x context for the task itself, or would you rather have core expose a
> > helper for that?
> 
> I think that we could reuse `io.quarkus.vertx.VertxContextSupport.executeBlocking()` helper in the tasks module.

Actually, I think that we should follow the logic we use for programmatic features, i.e. allow users to specify a handler with `ExecutionModel#WORKER_THREAD`, `ExecutionModel#EVENT_LOOP` or `ExecutionModel#VIRTUAL_THREAD`. Something like we do in `io.quarkiverse.mcp.server.runtime.FeatureManagerBase.execute(ExecutionModel executionModel, Callable<Uni<RESULT>> action)`.

PR #1014, opened 2026-09-21
Update spec-version to 2026-07-28

kdubois:
I believe since 2.0.0 we're compliant with the latest MCP spec version.


mkouba (approved):

mkouba:
Good catch, thanks!

